PR #13, opened 2016-09-15
Remove `sleep` and increase `wait` in tests

zda:
For me this change reduces the test suite time by 19%. More details on Mantis shortly.

### DIFF
--- a/test/spec/app_spec_helpers.rb
+++ b/test/spec/app_spec_helpers.rb
@@ -4,8 +4,7 @@ module AppSpecHelpers
   end
 
   def get_element(selector, type = :css)
-    sleep 3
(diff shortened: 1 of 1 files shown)


infojunkie:
It would be great to remove *all* `sleep` invocations and replace with `wait`.

zda:
@infojunkie, I agree that would be great. I created a ticket (#5124) to try to advance that goal, and hopefully this branch inspires us towards that end by demonstrating the possible savings.

That ticket is unscheduled however and thus I suspect it will not be completed in the immediate future. In the meantime, we have to run these tests frequently, and I don't see we would not merge a three-line change that shortens the time needed to run tests by ~20% immediately.

PR #18, opened 2016-10-07
Feature/4997 client permissions

zda:
This continues down the list of components on #4997, with hiding the media status dropdown (7ec9cf1). 

I had to extend Caio's new <Can> component to be able to call it from non-JSX (3ebb04b) and we should probably get his feedback on Monday before merging.


caiosba:
Code looks good. I'll find out what's up with Travis before merging.

zda:
For now I am merging these one by one to make it easier to determine which are failing and fix them.

PR #27, opened 2016-11-05
Feature/electionland checklist

unthinkingly:
WHAT

This adds the electionland checklist to the media template so it appears above the verification timeline on the media detail page.

BEFORE
screen shot 2016-11-04 at 6 14 01 pm

(No Checklist)

AFTER
screen shot 2016-11-04 at 3 43 04 pm

(There is a checklist and the header has a little more padding on top, and the media is 100px wider)


unthinkingly:
This ideally would be modified to appear only on the electionland team.

PR #44, opened 2017-01-05
CR: WIP proof of concept of browser testing with Sauce

zda:
For this to work, you need to:

- download Sauce Connect proxy (the Dockerfile might not work)
- run `npm run test:proxy` and wait for "start your tests"
- add `sauce_username` and `sauce_access_key` to config.yml
- ensure that client and API have same root domain in configs
- run `npm run test:browsers`

NOTE: I don't want to merge this right now, just making a PR to make it easier to leave CR comments.


Dockerfile
```diff
@@ -22,6 +22,13 @@ RUN cd /app && npm install
 COPY test/Gemfile test/Gemfile.lock /app/test/
 RUN cd /app/test && gem install bundler && bundle install --jobs 20 --retry 5
 
+# Sauce Labs proxy
+RUN mkdir -p /app/test/spec/sauce
+RUN cd /app/test/spec/sauce && curl -O https://saucelabs.com/downloads/sc-4.4.2-linux.tar.gz
+# TODO: throw error if `openssl sha1 sc-4.4.2-linux.tar.gz` != '57a07a14c5d95d72b6606ba34fceaf5bf76c2865'
+RUN pwd
+RUN cd /app/test/spec/sauce && tar -zxvf sc-4.4.2-linux.tar.gz
+
```
zda: I thought this would work but it didn't seem to last I tried, and I had to run this commands manually inside the container.

I'm not sure what the best location is – it doesn't need to be accessible from outside the container (and needs to be inside to work), so maybe a directory other than /app...

@bfoote if you know how to do line 28 that would be helpful :)

package.json
```diff
@@ -18,8 +18,10 @@
     "watch:styles": "watch 'npm run style' ./src/app/styles",
     "publish": "node scripts/server.js",
     "test": "npm run test:unit && npm run test:integration",
-    "test:integration": "cd test && rspec && cd ..",
     "test:unit": "cd test/unit && NODE_PATH=$NODE_PATH:. BABEL_ENV=production mocha setup.js *.spec.js",
+    "test:integration": "cd test && rspec spec/integration_spec.rb && cd ..",
+    "test:proxy": "cd test && ruby spec/sauce/sauce_connect.rb && cd ..",
```
zda: This command should be run first and separately from test:browsers (assuming you want Sauce to be able to reach your Docker check-app services), and will tail in the window you run it in. I tried automating this but you probably want to keep it running and then run the test command multiple times, so separate seems better.

package.json
```diff
@@ -18,8 +18,10 @@
...
+    "test:browsers": "cd test && rspec spec/browser_spec.rb && cd ..",
```
zda: I kinda think a single `rspec` for all browsers is probably worse than looping through browsers somehow and calling `rspec` once on each...

test/spec/app_spec.rb
```diff
@@ -779,3 +788,4 @@
     # end
   end
 end
+end
```
zda: Possibly a better pattern than a lambda, but wrapping rspec blocks in a class seemed more complex so I'm starting here.

Also, we can indent once we like this; it screwed up the diff though.

test/spec/browser_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+browsers = [
+  :chrome,
+  {
+    appiumVersion: '1.6.3',
+    deviceName: 'iPad Simulator',
+    platformName: 'iOS',
+    platformVersion: '10.0',
+    browserName: 'Safari',
+    javascriptEnabled: true
+  }
```
zda: Starting with an iPad because it's easier to debug due to the larger screen. Once it passes let's add iPhone.

test/spec/browser_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'yaml'
+require_relative './app_spec.rb'
+
+config = YAML.load_file('config.yml')
+webdriver_url = "http://#{config['sauce_username']}:#{config['sauce_access_key']}@ondemand.saucelabs.com:80/wd/hub"
+
+browsers = [
+  :chrome,
```
infojunkie: This key name is misleading. Shouldn't it be something like `safari_ios`?

zda: It's not a key name, just an entry in the array :) Sauce (Selenium?) expands it to {browserName: :chrome}.

An improvement(...hopefully) I'm working on now should be more clear – instead of an array, a hash that maps slugs (of our invention) to Selenium browser capability hash, like:

```
browser_definitions = {
  safari_ipad_ios10: {
    browserName: 'Safari',
    deviceName: 'iPad Simulator',
    platformName: 'iOS',
    platformVersion: '10.0',
    javascriptEnabled: true,
    appiumVersion: '1.6.3'
  },

  chrome54_macos1012: {
    browserName: 'chrome',
    version: '54.0',
    platform: 'macOS 10.12'
  },

  chrome54_windows10: {
    browserName: 'chrome',
    version: '54.0',
    platform: 'Windows 10'
  },

  chrome48_linux: {
    browserName: 'chrome',
    version: '48.0',
    platform: 'Linux'
  }
}
```

You could then run the spec against a subset of browsers similar to `run :chrome48_linux` and something would look up what `:chrome48_linux` means in Selenium terms as needed. Vaguely similar to what they're doing in this example repo except in rakefile form: 

https://github.com/saucelabs-sample-test-frameworks/Ruby-RSpec-Selenium/blob/master/Rakefile

(The issue with their approach is all the tests get run but with no ability to track what happens to them...)

infojunkie: Yes, thanks, that's what I was expecting ^^

Dockerfile
bfoote: You tagged the wrong user.

infojunkie: lol sorry @bfoote we meant @bnfinet :-)

bnfinet: :-)
this should validate the SHA1 signature for the downloaded binary
```
ENV SAUCE_SHA1=57a07a14c5d95d72b6606ba34fceaf5bf76c2865
RUN echo "${SAUCE_SHA1}  sc-4.4.2-linux.tar.gz" | sha1sum -c -
```

PR #45, opened 2017-01-05
Feature/5222 refactor annotation

melsawy:
- Refactor annotation to use `ProjectMedia`
- Change project page and search page to list items based on `ProjectMedia`
- Change media URL `project/[id]/media/[ProjectMediaId]`
- Change create/update/delete mutations for annotations to use `ProjectMedia`


data/relay.json
```diff
@@ -11361,7 +11361,7 @@
               "defaultValue": null
             },
             {
-              "name": "information",
+              "name": "embed_data",
```
infojunkie: what's the difference between `embed_data` and `embed` ?

melsawy: embed_data:  hold overridden fields (i.e. {title: 'new title'})
embed: our method to return embed object (annotation)

src/app/components/media/CreateMedia.js
```diff
@@ -64,7 +64,7 @@ class CreateMedia extends Component {
     };
 
     const onSuccess = (response) => {
-      const rid = response.createMedia.media.dbid;
+      const rid = response.createMedia.media.pm_dbid;
```
caiosba: I think we should replace the `CreateMediaMutation` by a `CreateProjectMediaMutation`

PR #65, opened 2017-02-19
Feature/responsive project page

unthinkingly:
These are finishing touches on the project page #5711.

If I do a PR on develop it looks like a big merge conflict, so doing it on experimental instead. 

Not sure if we can make this in before deployment but it is a pretty conservative changeset, and provides extra stability on mobile.


infojunkie:
There's a conflict to be resolved here.

unthinkingly:
This is merged now.

PR #78, opened 2017-03-06
Workaround for Sass compilation bug

zda:
I find that with the old command only up to ~65535 bytes
can be piped from Sass to PostCSS. This uses a temp file.

@karim, re: comments, I just removed the stderr log file since it isn't really necessary, and made the .tmp path change. Didn't break Travis.


infojunkie:
Superseded by later pull request.

PR #118, opened 2017-06-06
Feature/5923 code refactoring bridge vs check

infojunkie:
Creating PR to add comments.


.gitattributes
```diff
@@ -3,5 +3,5 @@ travis.zip.enc merge=ours
 README.md merge=ours
 src/app/customHelpers.js merge=ours
 src/app/components/media/MediaStatus.js merge=ours
-test/spec/status_spec.rb merge=ours
```
infojunkie: do we still need that file?

caiosba: I will update `.gitattributes`

src/app/components/BrowserSupport.js
```diff
@@ -1,5 +1,17 @@
 import React, { Component, PropTypes } from 'react';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, defineMessages } from 'react-intl';
+import MappedMessage from './MappedMessage';
+
+const messages = defineMessages({
+  disclaimer: {
+    id: "browserSupport.message",
+    defaultMessage: "While in beta, Check is optimized for Google Chrome on desktop.",
```
infojunkie: this could have been done using {appName}

src/app/components/LoginContainer.js
```diff
@@ -1,39 +1,64 @@
...
+const messages = defineMessages({
+  appNameHuman: {
+    id: 'loginContainer.appNameHuman',
+    defaultMessage: 'Check',
+  },
+  bridge_appNameHuman: {
+    id: 'bridge.loginContainer.appNameHuman',
+    defaultMessage: 'Bridge',
+  },
+  agreeTerms: {
```
caiosba: Looks like we don't need this key anymore.

src/app/components/LoginContainer.js
```diff
@@ -1,39 +1,64 @@
...
+    id: 'agree.terms',
+    defaultMessage: 'By signing in, you agree to the Check {tosLink} and {ppLink}.',
+  },
+  loginSupport: {
```
caiosba: Looks like we don't need this key anymore.

src/app/components/UserPasswordReset.js
```diff
@@ -7,12 +7,17 @@ import FlatButton from 'material-ui/FlatButton';
 import TextField from 'material-ui/TextField';
 import ResetPasswordMutation from '../relay/ResetPasswordMutation';
 import CheckContext from '../CheckContext';
+import config from 'config';
 
 const messages = defineMessages({
   emailNotFound: {
```
infojunkie: We can remove this string

src/app/components/media/MediaParentComponent.js
```diff
@@ -0,0 +1,24 @@
...
+  render() {
+
+    const extraComponents = [
+      <Can permissions={this.props.media.permissions} permission="create Dynamic">
+        <Translation annotated={this.props.media} annotatedType="ProjectMedia" />
+      </Can>
+    ];
+
+    return config.appName === 'bridge' ? (
+      <MediaComponent {...this.props} extras={extraComponents} />
```
infojunkie: Do we need the `if bridge` condition here?

src/app/helpers.js
```diff
@@ -42,10 +43,10 @@ function getStatusStyle(status, property) {
 // Skip team name if `skipTeam` is true.
 // Skip `prefix |` if `prefix` empty.
 function pageTitle(prefix, skipTeam, team) {
-  let suffix = 'Check';
+  let suffix = capitalize(config.appName);
```
infojunkie: That's incorrect - either we are not using the function `pageTitle` anymore (which I assume is the case since we replaced it with the `<PageTitle>` component) or this function should use the localizable human string.

src/app/helpers.js
```diff
@@ -63,9 +64,19 @@ function truncateLength(text, length = 100) {
   return truncate(text, { length, separator: /,? +/, ellipsis: '…' });
 }
 
+function capitalize(text) {
```
infojunkie: Assuming the function `pageTitle` is no longer needed, `capitalize` should no longer be needed too.

localization/react-intl/src/app/components/LoginContainer.json
```diff
@@ -1,13 +1,29 @@
...
+  {
+    "id": "bridge.loginContainer.appNameHuman",
+    "defaultMessage": "Bridge"
   },
   {
     "id": "agree.terms",
     "defaultMessage": "By signing in, you agree to the Check {tosLink} and {ppLink}."
   },
   {
+    "id": "login.support",
```
infojunkie: this string is no longer needed

.gitattributes
infojunkie: what do we need `.gitattributes` for if we're no longer merging anything?

test/unit/config.js
```diff
@@ -1 +1,4 @@
-window.config = {}
+var config = window.config = {
+  appName: 'check',
```
infojunkie: this will need to be fixed when we merge back one last time into `bridge-web` - no need to do it now.

.gitattributes
caiosba: let's remove it and handle the Travis stuff separately (e.g., using one VM for Check and one VM for Bridge)

infojunkie: oh it was for Travis... yes and let's do this work when we merge back into `bridge-web` for the final unification

src/app/components/media/MediaParentComponent.js
caiosba: Not exactly on this line, but on this file yes, to tell that only Bridge has the `Translation` as an extra component (the condition could be used to define `extraComponents`)

src/app/helpers.js
caiosba: We shouldn't use the `pageTitle` function altogether anymore.

src/app/helpers.js
caiosba: Verified that it's only used in this file.

src/app/helpers.js
caiosba: Still used on src/app/components/project/Project.js line 63, in a redundant way... let's change it.

src/app/components/media/MediaParentComponent.js
infojunkie: OK let's leave this for #6085

amoedoamorim: Do you suggest any changes here?

caiosba: The alternative would be to set `extraComponents` based on the `appName` instead of checking for the `appName` when rendering.

infojunkie: Imo we leave as is until we fix the webpack build.

PR #131, opened 2017-06-26
Feature/5126 browser testing rebase

clarissacastella:


infojunkie:
Clarissa, why do we need the Browserstack logo image?

test/spec/browser_spec.rb
```diff
@@ -1,49 +1,19 @@
...
 end
 
 describe "browser" do
-  browsers_to_test.each do |browser_id|
-     describe "#{browser_id}" do
-      include_examples "app", webdriver_url, browser_definitions[browser_id]
+  browser_definitions.each do |browser_id|
+    describe "#{browser_id}" do
+      browser_id[1]["browserstack.debug"] = true
+      include_examples "app", config['webdriver_url'], browser_id[1]
```
infojunkie: Can we use the existing config key `chromedriver_url` or do we need this new one?

test/spec/browser_spec.rb
```diff
@@ -1,49 +1,19 @@
...
+  browser_definitions[browser[0]] = browser[1]
 end
 
 describe "browser" do
-  browsers_to_test.each do |browser_id|
-     describe "#{browser_id}" do
-      include_examples "app", webdriver_url, browser_definitions[browser_id]
+  browser_definitions.each do |browser_id|
+    describe "#{browser_id}" do
+      browser_id[1]["browserstack.debug"] = true
```
infojunkie: Can we put this config in the yml file directly?

test/spec/pages/login_page.rb
```diff
@@ -78,6 +78,12 @@ def toggle_form_mode
   end
 
   def confirm_email(email) # TODO: test real email confirmation flow
-    request_api('/test/confirm_user', { email: email })
+    if @config.key?('proxy')
+      addr = @config['self_url'].sub 'test.', 'check-api.test.'
```
infojunkie: can we change the config `self_url` directly instead of hacking it here?

test/spec/browser_spec.rb
clarissacastella: I think it is better to have two keys since we can run 
`docker-compose -f docker-test.yml run web.test npm run test:quick`
`docker-compose -f docker-test.yml run web.test npm run test:integration`
`docker-compose -f docker-test.yml run web.test npm run test:browsers`
without changing the config file.

If we use the same key, every time we want to change from `:browsers` to `quick`or `integration` we need to change the config file.

But yes, we can use the same key. But we need to change the config time every time we need to change from `:browsers` to other test.

test/spec/browser_spec.rb
clarissacastella: Yes we can. But it will be always true. Never changes.

test/spec/pages/login_page.rb
clarissacastella: No, we can't. It is just for this case in particular.
The alternative is to add a new key in the config file.

PR #147, opened 2017-08-23
Feature/6174 source logs

melsawy:


src/app/relay/CreateCommentMutation.js
```diff
@@ -11,8 +11,8 @@ class CreateCommentMutation extends Relay.Mutation {
   getFatQuery() {
     let query = '';
     switch (this.props.parent_type) {
-    case 'source':
-      query = Relay.QL`fragment on CreateCommentPayload { commentEdge, source { annotations } }`;
+    case 'project_source':
+      query = Relay.QL`fragment on CreateCommentPayload { commentEdge, project_source { source { log, log_count } } }`;
```
amoedoamorim: I'm not sure about this. Should `log` be under `source` or `project_source`?

src/app/relay/CreateTagMutation.js
```diff
@@ -18,7 +18,7 @@ class CreateTagMutation extends Relay.Mutation {
       query = Relay.QL`fragment on CreateTagPayload { tagEdge, project_media { log, tags, log_count } }`;
       break;
     case 'project_source':
-      query = Relay.QL`fragment on CreateTagPayload { tagEdge, project_source { tags } }`;
+      query = Relay.QL`fragment on CreateTagPayload { tagEdge, project_source { tags, source { log, log_count } } }`;
```
amoedoamorim: Same here. I'm not sure about this. Should `log` be under `source` or `project_source`?

src/app/components/source/Source.js
```diff
@@ -173,6 +173,105 @@ const SourceContainer = Relay.createContainer(SourceComponent, {
               }
             }
           },
+          log(first: 10000) {
```
amoedoamorim: Seems that this is under directly under `source`, not `project_source`. Shouldn't it be under `project_sorce`, just like it is for `project_media`?

src/app/components/source/SourceComponent.js
```diff
@@ -1054,7 +1053,7 @@ class SourceComponent extends Component {
 
           { !isEditing ?
             <div>
-              { this.state.showTab === 'annotation' ? <Annotations annotations={source.annotations.edges.slice().reverse()} annotated={source} annotatedType="Source" /> : null }
+              { this.state.showTab === 'annotation' ? <Annotations annotations={source.log.edges} annotated={this.props.source} annotatedType="ProjectSource" /> : null }
```
amoedoamorim: *annotations* prop is pointing to a `source` while  *annotated* is pointing to the `project_source` `(this.props.source)`

amoedoamorim: This might be confusing because the data is structured like this: `this.props.source.source`, where the parent is the `project_source` and the child is the `source`. I don't like this naming but I followed how it was like for `media`.

src/app/relay/CreateCommentMutation.js
melsawy: Should be related to source, checkout this note https://mantis.meedan.com/view.php?id=6174#c22350

src/app/components/source/Source.js
melsawy: no it should be under `source` to list all activities across all `project_source`s https://mantis.meedan.com/view.php?id=6174#c22367

src/app/components/source/SourceComponent.js
melsawy: Yes as I'd like to create annotation using  `ProjectSourceID` as `annotated_id` and `ProjectSource` as `annotated_type but need to list logs for sources.

PR #164, opened 2017-09-11
Feature/6218 skipped integration tests

clarissacastella:


test/spec/api_helpers.rb
```diff
@@ -53,7 +55,15 @@ def api_create_team_project_and_claim(quit = false, quote = 'Claim')
     data = api_create_team_and_project
     claim = request_api 'claim', { quote: quote, email: data[:user].email, team_id: data[:team].dbid, project_id: data[:project].dbid }
     @driver.quit if quit
-    claim
+    claim  
+  end
+
+  def api_create_team_and_project
```
caiosba: Why this method duplicates the one above? They have the same name

test/spec/app_spec.rb
```diff
@@ -830,6 +788,150 @@
...
+      page.go(@config['api_path'] + '/test/session?email='+@user_mail)
+      #As the group creator, go to the members page and approve the joining request.
+      page = TeamsPage.new(config: @config, driver: @driver).load
+          .disapprove_join_team(subdomain: @team1_slug)
+      @wait.until {
+        expect(@driver.page_source.include?('Requests to join')).to be(false)
+      }      
+    end
+
+    #TO DO PROBLEMS WITH MENU
```
infojunkie: `FIXME` not `TO DO`

test/spec/app_spec.rb
```diff
@@ -830,6 +788,150 @@
...
+
+    it "should delete member from team", bin4:true do
+      page = Page.new(config: @config, driver: @driver)
+      page.go(@config['api_path'] + '/test/session?email='+@user_mail)
+      #As the group creator, go to the members page and approve the joining request.
+      sleep 1
+      page = TeamsPage.new(config: @config, driver: @driver).load
+          .delete_user_team(subdomain: @team1_slug)
+
+      expect(@driver.find_elements(:xpath => '//button').length == 4).to be(true)
```
caiosba: Would be good to add an `expect` before a user is rejected, to make sure that something changed.

test/spec/app_spec.rb
```diff
@@ -1313,6 +1417,7 @@
       results.last.location_once_scrolled_into_view
       sleep 5
       results = @driver.find_elements(:css, '.medias__item')
+      p results.size
```
infojunkie: We should not need to print stuff.

test/spec/app_spec.rb
```diff
@@ -830,6 +788,150 @@
...
+      api_logout
+
+      @driver = new_driver(webdriver_url,browser_capabilities)
+      page = Page.new(config: @config, driver: @driver)
+      page.go(@config['api_path'] + '/test/session?email='+user2.email)
+      page = TeamsPage.new(config: @config, driver: @driver).load
+      sleep 3
+      @driver.find_element(:class,"header-actions__menu-toggle").click
+      sleep 3
+      expect(@driver.page_source.include?('Manage team')).to be(false)
```
caiosba: Here you test that "Manage team" is not displayed to a team owner but doesn't test if it is displayed to a team owner.

caiosba: Please also test if a team owner can actually see the "Manage team" option.

test/spec/app_spec.rb
```diff
@@ -830,6 +788,150 @@
...
+      page.go(@config['api_path'] + '/test/session?email='+utp[:user].email)
+      #As the group creator, go to the members page and approve the joining request.
+      page = TeamsPage.new(config: @config, driver: @driver).load
+          .approve_join_team(subdomain: utp[:team].slug)
+      go(@config['self_url'] + '/'+utp[:team].slug+'/project/'+utp[:project].dbid.to_s)
+      sleep 3
+      @driver.find_element(:class,"header-actions__menu-toggle").click
+      sleep 3
+      expect(@driver.page_source.include?('Edit project')).to be(true)
+      api_logout
```
caiosba: This whole process of adding a user to a team should be handled by the test API, to avoid repetition between tests. Please tell us what this test endpoint should do (for example, "given a team, create a user under that team with a given role") and someone can implement that on the backend.

test/spec/app_spec.rb
```diff
@@ -830,6 +788,150 @@
...
+
+      @driver = new_driver(webdriver_url,browser_capabilities)
+      page = Page.new(config: @config, driver: @driver)
+      page.go(@config['api_path'] + '/test/session?email='+user2.email)
+      go(@config['self_url'] + '/'+utp[:team].slug+'/project/'+utp[:project].dbid.to_s)
+      sleep 3
+      @driver.find_element(:class,"header-actions__menu-toggle").click
+      sleep 3
+      expect(@driver.page_source.include?('Edit project')).to be(false)
```
caiosba: Same thing as above: please test that someone with the right permissions can actually see the "Edit project" option.

test/spec/app_spec.rb
caiosba: Please remove this line.

test/spec/pages/source_page.rb
```diff
@@ -20,4 +20,10 @@ def title
   def avatar
     element('.source__avatar')
   end
+
+  def delete_annotation
```
caiosba: At least in the PR, you are not using this method anywhere. Please delete it if not needed.

test/spec/api_helpers.rb
clarissacastella: One was old, with not accepting team name as parameter. Deleted.

test/spec/app_spec.rb
clarissacastella: I fixed the problem in this version and didn't removed the comment. 
Now it is removed.

test/spec/app_spec.rb
clarissacastella: Both expects added to pages/team_page.rb

PR #179, opened 2017-10-04
#5961 Fix user avatar in source comments

amoedoamorim:


src/app/components/source/Source.js
```diff
@@ -152,20 +153,7 @@ const SourceContainer = Relay.createContainer(SourceComponent, {
...
-                    dbid,
-                    accounts(first: 10000) {
-                      edges {
-                        node {
-                          url
-                        }
-                      }
-                    }
-                  }
+                  ${userFragment}
```
caiosba: The `userFragment` asks for a lot of data, including many data of that user's source, the user's memberships, etc. Do we need all that stuff here? Just asking to avoid performance problems.

amoedoamorim: Hmm, not really, we really could optimize the query. I just tried to reuse the fragment to keep it even with other user queries in the app, but this one might be a bit too much.

amoedoamorim: Maybe we can try to make fragments more feature-specific instead of entity-specific. (i.e. userTooltipFragment rather than userFragment), just an idea...

caiosba: Right. In general the fragments should be smaller in order to be more reusable...

PR #180, opened 2017-10-05
#5962 Remove User Tooltips

amoedoamorim:


caiosba:
Alelas, most important is to remove `userFrament` from that query, remember?

PR #204, opened 2017-11-04
Test/stabilization2

clarissacastella:
Fixes in order to help tests in Develop stabilization.
Part of a work in progress.

### DIFF
--- a/test/spec/app_spec.rb
+++ b/test/spec/app_spec.rb
@@ -65,10 +65,10 @@ shared_examples 'app' do |webdriver_url, browser_capabilities|
 
   # The tests themselves start here
   context "web" do
-    
+
     include_examples "custom"
 
-    it "should filter by medias or sources", bin3: true do
+    it "should filter by medias or sources", bin6: true do
       api_create_team_project_and_link 'https://twitter.com/TheWho/status/890135323216367616'
       @driver.navigate.to @config['self_url']
       wait_for_selector("card-with-border", :class)
--- a/test/spec/app_spec.rb
+++ b/test/spec/app_spec.rb
@@ -848,22 +848,26 @@ end
     #   skip("Needs to be implemented")
     # end
 
+
     it "should navigate between teams", bin5: true, quick: true do
       # setup
       @user_mail = "test" +Time.now.to_i.to_s+rand(9999).to_s + @user_mail
       @team1_slug = 'team1'+Time.now.to_i.to_s+rand(9999).to_s
-
       user = api_register_and_login_with_email(email: @user_mail, password: @password)
       team = request_api 'team', { name: 'Team 1', email: user.email, slug: @team1_slug }
       request_api 'project', { title: 'Team 1 Project', team_id: team.dbid }
       team = request_api 'team', { name: 'Team 2', email: user.email, slug: "team-2-#{rand(9999)}#{Time.now.to_i}" }
       request_api 'project', { title: 'Team 2 Project', team_id: team.dbid }
-      page = TeamsPage.new(config: @config, driver: @driver).load.select_team(name: 'Team 1')
+
+      page = MePage.new(config: @config, driver: @driver).load.select_team(name: 'Team 1')
+      page.click_projects_tab
+
       expect(page.team_name).to eq('Team 1')
       expect(page.project_titles.include?('Team 1 Project')).to be(true)
       expect(page.project_titles.include?('Team 2 Project')).to be(false)
 
-      page = TeamsPage.new(config: @config, driver: page.driver).load.select_team(name: 'Team 2')
+      page = MePage.new(config: @config, driver: @driver).load.select_team(name: 'Team 2')
+      page.click_projects_tab
 
       expect(page.team_name).to eq('Team 2')
       expect(page.project_titles.include?('Team 2 Project')).to be(true)
@@ -871,7 +875,7 @@ end
     
       #As a different user, request to join one team and be accepted.
       user = api_register_and_login_with_email(email: "new"+@user_mail, password: @password)
-      page = TeamsPage.new(config: @config, driver: @driver).load
+      page = MePage.new(config: @config, driver: @driver).load
       page.ask_join_team(subdomain: @team1_slug)
       @wait.until {
         expect(@driver.find_element(:class, "message").nil?).to be(false)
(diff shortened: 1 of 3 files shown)


test/spec/app_spec.rb
```diff
@@ -632,63 +619,58 @@
...
+      old = @driver.find_elements(:class,"annotations__list-item").length
 
       # Validate assumption that tag does not exist
       expect(media_pg.has_tag?(new_tag)).to be(false)
 
       # Try to add from command line
       media_pg.add_annotation("/tag #{new_tag}")
-      Selenium::WebDriver::Wait.new(timeout: 10).until { media_pg.has_tag?(new_tag) } # TODO: wait inside MediaPage
+      old = wait_for_size_change(old, "annotations__list-item", :class)
+      media_pg.has_tag?(new_tag)
```
caiosba: This line should be removed

test/spec/app_spec.rb
```diff
@@ -632,63 +619,58 @@
...
       expect(media_pg.has_tag?(new_tag)).to be(true)
 
       # Try to add duplicate from command line
       media_pg.add_annotation("/tag #{new_tag}")
 
       # Verify that tag is not added and that error message is displayed
       expect(media_pg.tags.count(new_tag)).to be(1)
+      sleep 3
```
caiosba: Can't we wait for the `.message` element here instead of `sleep`ing?

test/spec/app_spec.rb
```diff
@@ -632,63 +619,58 @@
...
 
-    it "should not create duplicated media", bin1: true do
+    it "should not create duplicated media", bin4: true do
       api_create_team_project_and_link_and_redirect_to_media_page @media_url
       id1 = @driver.current_url.to_s.gsub(/^.*\/media\//, '').to_i
       expect(id1 > 0).to be(true)
       @driver.navigate.to @driver.current_url.to_s.gsub(/\/media\/[0-9]+$/, '')
-
       sleep 3
+      wait_for_selector("medias__item",:class)
```
caiosba: Is it really needed to have a `sleep` here since it's followed by a `wait_for_selector`?

test/spec/app_spec.rb
clarissacastella: Agree

test/spec/app_spec.rb
```diff
@@ -632,63 +619,58 @@
...
       sleep 2
       press_button('#create-media-submit')
-      sleep 10
-
-      id2 = @driver.current_url.to_s.gsub(/^.*\/media\//, '').to_i
-      expect(id2 > 0).to be(true)
-
-      expect(id1 == id2).to be(true)
+      sleep 3
+      wait_for_selector("add-annotation__insert-photo",:class)
```
caiosba: Same question as above: Is it really needed to have a `sleep` here since it's followed by a `wait_for_selector`?

test/spec/app_spec.rb
clarissacastella: When the program loads a new page, that can be in blank in the beginning of the loading process, my experience views that it is prudent to have a small sleep before the wait_for_selector, that ensures that the waiting time is sufficient.

test/spec/app_spec.rb
```diff
@@ -632,63 +619,58 @@
...
-    it "should not create duplicated media", bin1: true do
+    it "should not create duplicated media", bin4: true do
       api_create_team_project_and_link_and_redirect_to_media_page @media_url
       id1 = @driver.current_url.to_s.gsub(/^.*\/media\//, '').to_i
       expect(id1 > 0).to be(true)
       @driver.navigate.to @driver.current_url.to_s.gsub(/\/media\/[0-9]+$/, '')
-
       sleep 3
+      wait_for_selector("medias__item",:class)
+      media_url = @driver.current_url.to_s.gsub(/\/media\/[0-9]+$/, '')
```
caiosba: This is actually the project URL, no?

test/spec/app_spec.rb
clarissacastella: Good idea. Will try it.

test/spec/app_spec.rb
clarissacastella: No. will fix it.

test/spec/app_spec.rb
clarissacastella: yes

test/spec/app_spec.rb
```diff
@@ -632,63 +619,58 @@
...
       press_button('#create-media-submit')
-      sleep 10
-
-      id2 = @driver.current_url.to_s.gsub(/^.*\/media\//, '').to_i
-      expect(id2 > 0).to be(true)
-
-      expect(id1 == id2).to be(true)
+      sleep 3
+      wait_for_selector("add-annotation__insert-photo",:class)
+      expect(@driver.current_url.to_s.gsub(/\/media\/[0-9]+$/, '') == media_url).to be(true)
```
caiosba: Here you are asserting that the user was redirected to a media page whose project is the same as the first media, or that the user didn't even leave the project page. I think we should keep the same assertion as before: the user was redirected to a media page which is the same as the existing media.

test/spec/app_spec.rb
```diff
@@ -832,28 +790,102 @@
     #   skip("Needs to be implemented")
     # end
 
-    it "should navigate between teams", bin4: true, quick: true do
+
+    it "should navigate between teams", bin5: true, quick: true do
```
caiosba: Just a suggestion: I think this test is way too big. We are actually testing switch teams, join team, approve user, delete user and reject user. I suggest to split into five or at least change the test title to reflect what's actually being tested here.

test/spec/check/custom_spec.rb
```diff
@@ -27,46 +27,35 @@
...
-  #   sleep 5
+    # Add a status as a command
+    fill_field('#cmd-input', '/status In Progress')
+    @driver.action.send_keys(:enter).perform
+    sleep 5
 
-  #   # Verify that status was added to annotations list
-  #   expect(@driver.page_source.include?('Status')).to be(true)
+    # Verify that status was added to annotations list
+    expect(@driver.page_source.include?('Status')).to be(true)
```
caiosba: Please add a `expect(@driver.page_source.include?('Status')).to be(false)` before the `fill_field` call above - this way we know that something actually changed.

test/spec/app_spec.rb
clarissacastella: Those tests are so connected that in order to avoid repetition I choose to join them and I think it worked fine.
I completely agree with the title suggestion.

test/spec/check/custom_spec.rb
clarissacastella: Agree

test/spec/app_spec.rb
clarissacastella: Ok

test/spec/app_spec.rb
clarissacastella: Using the analyzer I didn't find a name for the message code, like ".message". 
So I was not able to insert a code to wait for it, since the div/class has no name.

clarissacastella:
@caiosba could you review the changes?

PR #206, opened 2017-11-07
Test/stabilization3

clarissacastella:


clarissacastella:
+3 test methods.

test/spec/app_spec.rb
```diff
@@ -457,48 +457,49 @@
...
-      sleep 1
-      @driver.find_element(:class, 'source__edit-save-button').click
-      sleep 5
+      list = wait_for_selector_list("svg[class='create-task__remove-option-button create-task__md-icon']")
+      str= @driver.page_source
+      str = str[str.index('www.acdc.com')..str.length]
+      str = str[str.index('source__remove-link-button')..str.length]
+      str = str[0..str.index('tabindex') - 3]
+      str = "." +str.gsub(" ", ".")
+      element = wait_for_selector_list(str)[1]
```
caiosba: This whole block starting on line 490 seems complex... what's the idea of it?

test/spec/app_spec.rb
```diff
@@ -457,48 +457,49 @@
...
-    it "should add and remove accounts to sources", bin6: true  do
+    it "should add and remove accounts to sources", bin6: true do
       api_create_team_project_and_source_and_redirect_to_source('GOT', 'https://twitter.com/GameOfThrones')
-      #sleep 5
-      element = wait_for_selector("source__edit-button", :class)
+      wait_for_selector("source__tab-button-account",:class)
+      element = wait_for_selector("source__edit-button",:class)
       element.click
-      #sleep 3
+      sleep 1
```
infojunkie: do we still need the `sleep`s in these tests?

test/spec/app_spec.rb
```diff
@@ -767,13 +768,26 @@
...
+      el.click
+      el = wait_for_selector('.media-actions__edit')
+      el.click
+      wait_for_selector('ReactTags__selected', :class)
+      fill_field('.ReactTags__tagInput input', "TAG")
+      @driver.action.send_keys("\n").perform      
+      sleep 1
+      fill_field('.ReactTags__tagInput input', "TAG2")
+      @driver.action.send_keys("\n").perform
+      expect(page.tags.length == 2).to be(true)
```
caiosba: Please make sure that the number of tags is 0 before the first tag is added.

test/spec/app_spec.rb
```diff
@@ -783,10 +797,51 @@
...
+    it "should edit team and logo", bin1: true do
+      team = "testteam#{Time.now.to_i}"
+      api_create_team(team:team)
+      p = Page.new(config: @config, driver: @driver)
+      p.go(@config['self_url'] + '/' + team)
+      wait_for_selector("team-members__edit-button", :class)
+      str= @driver.page_source
+      str = str[str.index('team__edit-button')..str.length]
+      str = str[0..str.index('tabindex') - 3]
+      str = "." +str.gsub(" ", ".")
```
caiosba: Same question as before: what's the idea of this block that goes from line 806 until here?

test/spec/app_spec.rb
```diff
@@ -783,10 +797,51 @@
...
+      el = wait_for_selector("team__link-container", :id)
+      el.click
+      el.send_keys "www.meedan.com"
+
+      #Logo
+      str= @driver.page_source
+      str = str[str.index('team__edit-avatar-button')..str.length]
+      str = str[0..str.index('>') - 2]
+      str = "." +str.gsub(" ", ".")
+      el = wait_for_selector(str)
```
caiosba: Same as before: what's the idea of this block?

test/spec/app_spec.rb
```diff
@@ -783,10 +797,51 @@
...
+      el.click
+      el.send_keys "EDIT"
+
+      el = wait_for_selector("team__description-container", :id)
+      el.click
+      el.send_keys "EDIT DESCRIPTION"
+
+      el = wait_for_selector("team__location-container", :id)
+      el.click
+      el.send_keys "EDIT DESCRIPTION"
```
caiosba: Should be "location", not "DESCRIPTION".

test/spec/app_spec.rb
```diff
@@ -783,10 +797,51 @@
...
+      str= @driver.page_source
+      str = str[str.index('team__edit-avatar-button')..str.length]
+      str = str[0..str.index('>') - 2]
+      str = "." +str.gsub(" ", ".")
+      el = wait_for_selector(str)
+      el.click
+      el = wait_for_selector("team__save-button", :class)
+      el.click
+      wait_for_selector("team-members__edit-button", :class)
+      expect(@driver.page_source.include?('Team information updated successfully!')).to be(true)
```
caiosba: Please make sure that the message is not present before the "save" button clicked. Also, I suggest to add a few more `expect` here to verify that the updated information is displayed on the page. For example, `expect(@driver.page_source.include?('EDIT DESCRIPTION')).to be(false)` before and `expect(@driver.page_source.include?('EDIT DESCRIPTION')).to be(true)` after.

test/spec/app_spec.rb
clarissacastella: The idea is getting the name of the class.
Because the button id is "source__remove-link-button sc-jTzLTM dtwjjF "
And there are more than one delete button like this
And they don't have any text to use to find it.

infojunkie: As previously discussed, plz coordinate with @amoedoamorim to add classes where needed.

test/spec/app_spec.rb
clarissacastella: Because between click in the link to add and start to insert the information the browser lost it self if it is immediate .

test/spec/app_spec.rb
clarissacastella: Ok. Agree

test/spec/app_spec.rb
clarissacastella: Same answer as before.

test/spec/app_spec.rb
clarissacastella: Same answer as before.
Get the name of a button without text or unique id.

test/spec/app_spec.rb
clarissacastella: Agree

test/spec/app_spec.rb
clarissacastella: Ok

test/spec/app_spec.rb
caiosba: The `class` value is `source__remove-link-button sc-jTzLTM dtwjjF` which means that it has three classes `source__remove-link-button`, `sc-jTzLTM` and `dtwjjF`. You can look for class `source__remove-link-button` and, if there is more than one, you can use its position in the array. Makes sense?

infojunkie: sorry, but I do think it's better to place human class names, rather than assume button positions.

test/spec/app_spec.rb
clarissacastella: I made a new test removing the sleeps and fails :(

test/spec/app_spec.rb
caiosba: But `source__remove-link-button` is human, no? I mean, if there is more than one element with the same human class name, use their positions on the source to differentiate between them.

clarissacastella: Worked!!
Thank you.

PR #217, opened 2017-11-17
Test/stabilization5

clarissacastella:
Please review.

### DIFF
--- a/test/spec/app_spec.rb
+++ b/test/spec/app_spec.rb
@@ -1138,9 +1138,62 @@ shared_examples 'app' do |webdriver_url, browser_capabilities|
       delete_task('Foo')
     end
 
-    # it "should add, edit, answer, update answer and delete single_choice task" do
-    #   skip("Needs to be implemented")
-    # end
+    it "should add, edit, answer, update answer and delete single_choice task", bin2: true  do
+      media_pg = api_create_team_project_and_claim_and_redirect_to_media_page
+      wait_for_selector('.create-task__add-button')
+      # Create a task
+      expect(@driver.page_source.include?('Foo or bar?')).to be(false)
+      expect(@driver.page_source.include?('Task "Foo or bar?" created by')).to be(false)
+      el = wait_for_selector('.create-task__add-button', :css)
+      el.click
+      sleep 1
+      el = wait_for_selector('create-task__add-choose-one', :class)
+      el.click
+      wait_for_selector('#task-label-input', :css)
+      fill_field('#task-label-input', 'Foo or bar?')
+      fill_field('0', 'Foo', :id)
+      fill_field('1', 'Bar', :id)
+      el = wait_for_selector('.create-task__dialog-submit-button', :css)
+      el.click
+      media_pg.wait_all_elements(2, "annotations__list-item", :class) #Wait for refresh page
+      expect(@driver.page_source.include?('Foo or bar?')).to be(true)
+      expect(@driver.page_source.include?('Task "Foo or bar?" created by')).to be(true)
+      # Answer task
+      expect(@driver.page_source.include?('Task "Foo or bar?" answered by')).to be(false)
+      el = wait_for_selector('0', :id)
+      el.click
+      el = wait_for_selector('task__submit', :class)
+      el.click
+      media_pg.wait_all_elements(3, "annotations__list-item", :class)
+      expect(@driver.page_source.include?('Task "Foo or bar?" answered by')).to be(true)
+      # Edit task
+      expect(@driver.page_source.include?('Task "Foo or bar?" edited to "Foo or bar???" by')).to be(false)
+      el = wait_for_selector('.task-actions__icon', :css)
+      el.click
+      media_pg.wait_all_elements(6, "annotations__list-item", :class)
+      editbutton = wait_for_selector('.task-actions__edit', :css)
+      editbutton.location_once_scrolled_into_view
+      editbutton.click
+      fill_field('textarea[name="label"]', '??')
+      editbutton = wait_for_selector('.task__save', :css)
+      editbutton.click
+      media_pg.wait_all_elements(7, "annotations__list-item", :class)
+      expect(@driver.page_source.include?('Task "Foo or bar?" edited to "Foo or bar???" by')).to be(true)
+      # Edit task answer
+      expect(@driver.page_source.gsub(/<\/?[^>]*>/, '').include?('Task "Foo or bar???" answered by User With Email: "Foo edited"')).to be(false)
+      el = wait_for_selector('.task-actions__icon', :css)
+      el.click
+      el = wait_for_selector('.task-actions__edit-response', :css)
+      el.click
+      el = wait_for_selector('1', :id)
+      el.click
+      el = wait_for_selector('task__submit', :class)
+      el.click
+      media_pg.wait_all_elements(8, "annotations__list-item", :class) #Wait for refresh page
+      expect(@driver.page_source.gsub(/<\/?[^>]*>/, '').include?('Task "Foo or bar???" answered by User With Email: "Bar"')).to be(true)
+      # Delete task
+      delete_task('Foo')    
+    end
 
     # it "should add, edit, answer, update answer and delete multiple_choice task" do
     #   skip("Needs to be implemented")
@@ -1414,7 +1467,7 @@ shared_examples 'app' do |webdriver_url, browser_capabilities|
     end
 =end
 
-    it "should add image to media comment", bin: true, bin3: true do
+    it "should add image to media comment", bin3: true do
       api_create_team_project_and_claim_and_redirect_to_media_page
       # First, verify that there isn't any comment with image
       expect(@driver.page_source.include?('This is my comment with image')).to be(false)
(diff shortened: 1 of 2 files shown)


test/spec/app_spec.rb
```diff
@@ -585,28 +580,34 @@
...
       @driver.action.send_keys(:down).perform
       @driver.action.send_keys(:return).perform
-      sleep 2
+      sleep 1
       @driver.navigate.refresh
-      sleep 3
+      sleep 2
+      wait_for_size_change(0, "sourceLanguageInput",:id)
+      wait_for_selector("source__tab-button-media",:class)      
       expect(@driver.page_source.include?('Acoli')).to be(true)
```
caiosba: Shouldn't this be asserted before the `@driver.navigate.refresh` in order to be sure that auto-refresh worked?

test/spec/app_spec.rb
```diff
@@ -585,28 +580,34 @@
...
-      list = @driver.find_elements(:css => "div.source-tags__tag svg")
-      list[0].click
+      elements =wait_for_selector_list("div.source-tags__tag svg")
+      elements[0].click
       sleep 1
       @driver.navigate.refresh
-      sleep 3
+      sleep 2
+      wait_for_selector("source__tab-button-media",:class)
       expect(@driver.page_source.include?('Acoli')).to be(false)
```
caiosba: Same as above. Assert that the language was removed even if the page was not refreshed.

test/spec/app_spec.rb
```diff
@@ -1038,11 +1018,32 @@
       expect(@driver.page_source.include?('Auto-Refresh')).to be(true)
     end
 
-    # it "should cancel request through switch teams" do
-    #   skip("Needs to be implemented")
-    # end
+    it "should give 404 when trying to acess a media that is not related to the project on the URL", bin1: true do
```
caiosba: Typo: `acess` -> `access`

test/spec/app_spec.rb
```diff
@@ -1038,11 +1018,32 @@
...
+      el.click
+
+      fill_field('team-name-container', team, :id)
+      el = wait_for_selector('team-slug-container',:id)
+      el.click
+      el.clear
+      fill_field('team-slug-container', team, :id)
+      el = wait_for_selector("create-team__submit-button",:class)
+      el.click
+      wait_for_selector("team-members__profile-link",:class)
```
caiosba: From line 1023 until here: why do you need to create a team using the web UI?

test/spec/app_spec.rb
```diff
@@ -1038,11 +1018,32 @@
...
+      url = @config['self_url'] + '/'+ team +'/project/'+data.project_id.to_s + '/media/'+data.media_id.to_s
```
caiosba: You tested something different: you tested that the media is not related to the team, not the project. I suggest that you just call `api_create_team_project_and_claim_and_redirect_to_media_page` (which is faster than a link and you don't need necessarily a link here), get the current URL, replace the project id in the URL by `0` and assert that you reach a "Not Found" page.

test/spec/app_spec.rb
```diff
@@ -1290,7 +1344,7 @@
         expect(@driver.page_source.include?('Embed')).to be(false)
         @driver.navigate.to "#{@driver.current_url}/embed"
         sleep 2
-        expect(@driver.page_source.include?('Not available')).to be(true)
+        expect(@drivre.page_source.include?('Not available')).to be(true)
```
caiosba: Should be `driver`

test/spec/app_spec.rb
```diff
@@ -1137,9 +1138,62 @@
...
+      el = wait_for_selector('.task-actions__edit-response', :css)
+      el.click
+      el = wait_for_selector('1', :id)
+      el.click
+      el = wait_for_selector('task__submit', :class)
+      el.click
+      media_pg.wait_all_elements(8, "annotations__list-item", :class) #Wait for refresh page
+      expect(@driver.page_source.gsub(/<\/?[^>]*>/, '').include?('Task "Foo or bar???" answered by User With Email: "Bar"')).to be(true)
+      # Delete task
+      delete_task('Foo')
```
caiosba: Does this method assert that the task was actually deleted?

test/spec/app_spec.rb
clarissacastella: I think the result is the same, `wait_for_size_change(0, "sourceLanguageInput",:id)` being before or after the refresh. But I'll change.

The `wait_for_selector("source__tab-button-media",:class) ` needs to be after, it ensures that the page is completely loaded.

test/spec/app_spec.rb
clarissacastella: The `wait_for_selector("source__tab-button-media",:class) ` needs to be after refresh, it ensures that the page is completely loaded.

test/spec/app_spec.rb
clarissacastella: :+1:

test/spec/app_spec.rb
caiosba: I mean the line `expect(@driver.page_source.include?('Acoli')).to be(true)`. You're checking after the `@driver.navigate.refresh`. We need to be sure that the language is added after the user presses `ENTER`. The user shouldn't need to refresh the whole page to see the new language he/she added.

test/spec/app_spec.rb
caiosba: Same as https://github.com/meedan/check-web/pull/217#discussion_r151712252

test/spec/app_spec.rb
clarissacastella: It is not necessary.
Updated to `      api_create_team(team: team)`

clarissacastella: Updated to `api_create_team_and_project()` indeed

test/spec/app_spec.rb
clarissacastella: Updated to use  `api_create_team_and_project()` + `api_create_team_project_and_link` to use a real project id instead of `0`

test/spec/app_spec.rb
clarissacastella: :+1:

test/spec/app_spec.rb
clarissacastella: I think it so.
https://github.com/meedan/check-web/blob/develop/test/spec/app_spec_helpers.rb#L81

clarissacastella: I think it so
https://github.com/meedan/check-web/blob/develop/test/spec/app_spec_helpers.rb#L81

caiosba: Right!

test/spec/app_spec.rb
```diff
@@ -585,28 +580,34 @@
...
       @driver.action.send_keys(:down).perform
       @driver.action.send_keys(:return).perform
-      sleep 2
+      sleep 1
+      wait_for_size_change(0, "sourceLanguageInput",:id)
       @driver.navigate.refresh
-      sleep 3
+      sleep 2
+      wait_for_selector("source-tags__tag",:class)  
       expect(@driver.page_source.include?('Acoli')).to be(true)
```
caiosba: Shouldn't we copy (not move) the line `expect(@driver.page_source.include?('Acoli')).to be(true)` to after line 597, before the `@driver.navigate.refresh`?

test/spec/app_spec.rb
```diff
@@ -585,28 +580,34 @@
...
-      list = @driver.find_elements(:css => "div.source-tags__tag svg")
-      list[0].click
+      elements =wait_for_selector_list("div.source-tags__tag svg")
+      elements[0].click
       sleep 1
       @driver.navigate.refresh
-      sleep 3
+      sleep 2
+      wait_for_selector("source__tab-button-media",:class)
       expect(@driver.page_source.include?('Acoli')).to be(false)
```
caiosba: Same as above... shouldn't we copy `expect(@driver.page_source.include?('Acoli')).to be(false)` to after line 607 (before `@driver.navigate.refresh`) in order to be sure that the language was removed without needing to refresh the page?

test/spec/app_spec.rb
clarissacastella: The idea is to see if the tag with the language is being viewed in the page not in the edit mode.

test/spec/app_spec.rb
clarissacastella: The idea is to see if the tag with the language was deleted in the page not in the edit mode.

test/spec/app_spec.rb
caiosba: But after we press "Enter" we exit edit mode. I mean that we don't need to refresh the page to see that the item was actually saved. We can just press "Save". Here: https://www.youtube.com/watch?v=a7w4qDacORA

test/spec/app_spec.rb
caiosba: But after we press "Enter" we exit edit mode. I mean that we don't need to refresh the page to see that the item was actually saved. We can just press "Save". Here: https://www.youtube.com/watch?v=a7w4qDacORA

test/spec/app_spec.rb
clarissacastella: :+1: 
Replacing the refresh by press save

test/spec/app_spec.rb
clarissacastella: Replacing the refresh by press save

PR #221, opened 2017-11-28
Test/stabilization7

clarissacastella:
Please review

### DIFF
--- a/test/spec/app_spec.rb
+++ b/test/spec/app_spec.rb
@@ -1031,9 +1031,19 @@ shared_examples 'app' do |webdriver_url, browser_capabilities|
       expect(title.text == 'Not Found').to be(true)
     end
 
-    # it "should cancel request through switch teams" do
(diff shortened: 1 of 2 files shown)


test/spec/app_spec.rb
```diff
@@ -1031,21 +1026,44 @@
...
+    it "should find all medias with an empty search", bin1: true do
+      api_create_media_and_go_to_search_page
+      old = wait_for_selector_list("medias__item", :class).length
+      el = wait_for_selector("search-input", :id)
+      el.click
+      @driver.action.send_keys(:enter).perform
+      sleep 3 #due the reload
+      wait_for_selector("search-input", :id)
+      current = wait_for_selector_list("medias__item", :class).length
+      expect(old == current).to be(true)
```
caiosba: Would be good to check that, besides being equal, they are greater than zero.

clarissacastella: Implementing

PR #228, opened 2017-12-06
4996 lint

infojunkie:


infojunkie:
On visible issue so far: modifying and deleting source metadata does not work and fails silently.

PR #314, opened 2018-05-23
Feature/6767 locking an annotation

amoedoamorim:


src/app/components/media/Media.js
```diff
@@ -73,7 +73,6 @@ const MediaContainer = Relay.createContainer(MediaParentComponent, {
           id
           dbid
           locked
-          permissions
```
caiosba: You don't need permissions anymore?

amoedoamorim: I'm getting the permission value from the ProjectMedia object. Since we don't have different perm values for each annotation I think it's more economic to get it from the parent obj

PR #353, opened 2018-10-03
#7113 search help

infojunkie:


src/app/components/Search.js
```diff
@@ -62,6 +68,31 @@ const StyledSearchInput = styled.input`
...
+  }
+
+  a {
+    font: ${caption};
+    padding-left: ${units(1)};
+  }
+
+  button {
+    color: ${black54};
+    float: right;
```
caiosba: Should it be aligned on the right for Arabic too? If not, please use different values based on `props.isRtl` like above

src/app/components/Search.js
```diff
@@ -62,6 +68,31 @@ const StyledSearchInput = styled.input`
...
+    display: block;
+  }
+
+  td {
+    padding: ${units(1)};
+  }
+
+  a {
+    font: ${caption};
+    padding-left: ${units(1)};
```
caiosba: Please make sure you don't need `padding-right` for Arabic as well

PR #446, opened 2019-07-18
Feature/7571 two factor authentication

melsawy:


src/app/components/Login.js
```diff
@@ -384,6 +386,24 @@ class Login extends Component {
...
               </div>
 
+              {this.state.type === 'login' ?
+                <div className="login__otp_attempt">
+                  <TextField
+                    fullWidth
+                    name="otp_attempt"
+                    value={this.state.otp_attempt}
+                    className="login__otp_attempt-input"
+                    ref={(i) => { this.inputName = i; }}
```
amoedoamorim: looks like this ref was copied by mistake

src/app/components/source/UserSecurity.js
```diff
@@ -0,0 +1,522 @@
...
+
+    if (!currentUser || !user || currentUser.dbid !== user.dbid) {
+      return null;
+    }
+
+    const style = {
+      margin: `${units(2)} 0`,
+    };
+
+    const cardStyle = {
```
caiosba: This style is the same as above, can they be unified?

src/app/components/source/UserSecurity.js
```diff
@@ -0,0 +1,522 @@
...
+    const subTitleStyle = {
+      margin: '5px 0px',
+      textTransform: 'uppercase',
+    };
+
+    const divBackupStyle = {
+      lineHeight: '25px',
+      fontWeight: 'bold',
+      fontSize: '12px',
+      backgroundColor: '#dfdfe6',
```
caiosba: If you really need a new color, please add it to `src/app/styles/js/shared.js` and use it here... we usually avoid having hard-coded colors in the code

src/app/components/source/UserSecurity.js
```diff
@@ -0,0 +1,522 @@
...
+                      </h3>
+                      <span style={{ lineHeight: '25px' }}>
+                        {
+                          <FormattedMessage
+                            id="userSecurity.downloadDescription"
+                            defaultMessage="You'll need a two-factor app, like Google Authenticator, on your smartphone to proceed:"
+                          />
+                        }
+                      </span>
+                      <Link to={appsUrls.apple} target="_blank" style={{ padding: '5px' }} >
```
caiosba: Since this is a external URL, I think it's better to use `<a href>` directly instead of `<Link />`... and please add the `rel="noopener noreferrer"` attribute to it

src/app/components/source/UserSecurity.js
```diff
@@ -0,0 +1,522 @@
...
+    const loginTrial = 4;
+
+    return (
+      <div id="user__security">
+        <h2 style={style}>
+          <FormattedMessage id="userSecurity.notification" defaultMessage="Notification" />
+        </h2>
+        <Card style={cardStyle}>
+          <CardText style={cardTextStyle}>
+            <span style={{ minWidth: '500px', padding: '0px' }}>
```
amoedoamorim: We don't set sizes in pixels, instead we use `units()` for a 8px grid normalization. Please use `units()` instead of specifying sizes in pixels directly.

src/app/components/source/UserSecurity.js
```diff
@@ -0,0 +1,522 @@
...
+                      <Link to={appsUrls.apple} target="_blank" style={{ padding: '5px' }} >
+                        <img src={appsUrls.appleImage} alt="" />
+                      </Link>
+                      <Link to={appsUrls.play} target="_blank" style={{ padding: '5px' }}>
+                        <img src={appsUrls.playImage} alt="" />
+                      </Link>
+                    </CardText>
+                    <CardText style={cardTextAuthStyle}>
+                      <h3 style={subTitleStyle}>
+                        {
```
amoedoamorim: is this bracket needed?

src/app/components/source/UserSecurity.js
```diff
@@ -0,0 +1,522 @@
...
+                          <FormattedMessage
+                            id="userSecurity.qrcodeHeader"
+                            defaultMessage="Step 3: Scan"
+                          />
+                        }
+                      </h3>
+                      <span style={{ lineHeight: '25px' }}>
+                        {
```
amoedoamorim: are these brackets needed? seems there are a few of them

src/app/components/Login.js
melsawy: yes removed it

src/app/components/source/UserSecurity.js
melsawy: Fixed by removing one of these styles.

src/app/components/source/UserSecurity.js
melsawy: use existing one from src/app/styles/js/shared.js

src/app/components/source/UserSecurity.js
melsawy: Fixed

src/app/components/source/UserSecurity.js
melsawy: removed unneeded bracket

src/app/components/source/UserSecurity.js
melsawy: no I removed all unneeded brackets

src/app/components/source/UserSecurity.js
melsawy: Done

PR #496, opened 2019-12-11
471 hosam config runtime

huslage:


huslage:
@hfikry92 can you squash these commits please?

huslage:
@hfikry92 The `docker-entrypoint.sh` script is for local development and doesn't need the configurator to be applied inside the container. Also please remove the changes to `.gitlab-ci.yml`.

PR #550, opened 2020-03-01
Feature/67955 requests and comments in two distinct tabs

melsawy:


src/app/components/annotations/Annotation.js
```diff
@@ -1072,6 +1073,16 @@ class Annotation extends Component {
...
 
+      if (object.field_name === 'smooch_data' && activityType === 'create_dynamicannotationfield') {
+        showCard = true;
+        const messageText = JSON.parse(object.value).text;
+        contentTemplate = (
+          <div className="annotation__card-content">
+            <ParsedText text={messageText} />
+          </div>
+        );
+      }
```
caiosba: I think there are situations where there is no text (for example, if the user submits only an image)... what happens in that case? You render an empty card?

melsawy: Yes render an empty card, shall I hide the request if no text exist?
if yes how can owners delete these requests?
and also the number of requests will not be same as requests below requests tab

PR #610, opened 2020-05-12
Feature/8228 nix ag grid

adamhooper:
* [#8228] [#8090] [#8179] fixes error messages from ag-grid
* [#8246] checking an item no longer makes the page jump (unless
  paginator gets longer, causing wrapping)
* [#8249] uses material-ui Table

Reduces `vendor.bundle.js + index.bundle.js` from `5.27MB` to `4.30MB` (18% reduction). Should lead to more-than-18% page-load speedup, because we had to force ag-grid to _re-render_ just to make it work in React. [#8228].

The biggest change (aside from bugfixes) is in sizing and scrollbars. Now, the only scrollbars are the _page_ scrollbars, and the header sticks to the top of the page during scroll. (I had wanted to make more of the headers stick to the top of the page; but that turns out to be hard in 2020 because `<thead>` can't have `position: sticky`; and @pierreconti said this is okay.)

Minor tweaks (vetted by @pierreconti, though of course nothing is permanent):

* Item titles are centered vertically.
* Font size is the page-default font size
* Previously, each cell was a `<a>`, and all `<a>` pointed to the same URL. That meant the user needed to click on _text_ (as opposed to the _row_) to open the item. It also meant hovering over a cell would underline just that cell (as opposed to the _row_). Now, hovering a row changes the background (material-ui style), and clicking the _row_ changes the URL.
* Numbers are centered and formatted.
* Datetimes are rendered with react-intl's `<FormattedRelative>`. It's like `intl.formatRelative()` but it auto-updates at a tasteful interval to change "10 seconds ago" to "1 minute ago" as the user stays on the page.


adamhooper:
@caiosba @amoedoamorim Thank you for your previous comments. In addressing them, I found a rabbit hole: URLs. I have tweaked the table's CSS to match your comments and Pierre's, _and_ I revamped the media-page URLs.

First off, I broke `Search.js` into a new page, `AllItems.js`, This means `<AllItems>`, `<Trash>` and `<Project>` (all "pages", which receive router params) all call `<Search>`. Previously, `<Search>` had two jobs: to behave like `<AllItems>`, _and_ to behave like a sub-component of `<Trash>` and `<Project>`.

Then came a revamp of the Project-Media page, which I'll copy/paste here from Slack:

ProjectMedia URL query string can contain `listIndex`, `listQuery` and `listPath` -- but it only contains the ones it _needs_:

* `listIndex` is always used, for navigation. (If it's missing, no nav links appear.)
* `listQuery` is set if it there are filters/sorts, but it's unset by default.
* `listPath` is set if we're looking at `/trash` (... and someday maybe we could instead route to `/trash/media/:mediaId` and then we wouldn't need it.)

And the "Next"/"Previous" buttons? They do the search _before redirecting_. This fixes the browser's Back button. Click the "Next" button, and it turns into a spinner while running a GraphQL query with `listQuery @ listIndex+1`. When the query returns, the button executes `browserHistory.push` and gets unmounted.

All this means we no longer need `location.state`. URLs survive Refresh, Open in new Tab, or copy/paste.

It's joyous. The navbar never gets unmounted at all -- making a flicker-free experience for the user. I force the action buttons to re-mount Just To Be Safe, but they shouldn't need to unmount, either.

... And finally, a few odds and ends make it all work:

* `<BulkActions>`: `project` prop comes from props, never context.
* `<SearchResults>`: No more `CheckContext`
* `<CreateMedia>`: No more `CheckContext`
* `CreateProjectMediaMutation`: nix `props.context`, since we already know `team` (which is all `props.context` was passing).
* `optimisticProjectMedia`: change types (`linked_items_count: 0` instead of `linked_items_count: '0'`)
* Move `const muiTheme = createMuiTheme(...)` to `styles/js/shared.js`, so we don't re-create the theme every page transition.

PR #634, opened 2020-06-08
Warnings, cleanups [#8179] [#8222]

adamhooper:
* Integration tests: nix `appium` from Gemfile. The only thing it gave was `Selenium::WebDriver::Element#value` ... so replace that with `Selenium::WebDriver::Element#property('value')`. Nixes ~20 lines of warnings during integration-test run [#8222].
* Unit tests: nix all Enzyme `render()` calls. Use `mount()` and `shallow()`, only. (`render()` is for server-side rendering.). Nixes ~800 lines of warnings during unit-test run [#8179].
* Nix some `injectIntl()` [#8347].
* Nix some `isRtl` variables [#8249].
* `UploadImage`: behave like a controlled component [#8362]. This turned into a pretty big audit, and it led to these fixes:
    * `<LoginContainer>`: don't "toggle" between "login" and "register". Make a "login" button and a "register" button. (One step closer to removing state here. Fixes some integration-test flakiness.)
    * `<ReportDesigner>`: store only one `image` in state, not two. Don't save derived state, `hasUnsavedChanges`: compute it on demand instead.
    * `<ReportDesigner>`: after save, derive state from server response instead of assuming client-side state is up-to-date.
    * `<ReportImagePreview>`: avoid HTML injection. Use a single `image` prop. Switch to Hooks. (Had to revamp things when fixing `<UploadImage>`.)
* `<MoreLess>`: don't store derived state, `canExpand`. Store `contentHeight` in state and make `MaxHeight` a constant, instead. Make callers use `key` so state resets when text changes. (Fixes button misbehaving when navigating between two already-loaded media.) Also fixes unit-test warning [#8179].
* `<LoginContainer>`: Wrap with `<div>`, not `<p>` [#8179]


adamhooper:
@caiosba @amoedoamorim This turned into a pretty scary pull request! I can spend the time breaking it up, but if you're game I propose we try all in one go.

The TL;DR is: Travis output on this branch is 6,732 lines; on `develop` it's 8,490. That's a reduction of ~1,750 lines of warnings from automated tests!

@caiosba The scariest bit that touches _your_ work is `ReportDesigner`. I revamped `UploadImage`, and then the upload image _looked_ right, but I found that if I ran the steps "1. save a report design; 2. refresh the page; 3. look at state", the state was different in step 3 than in step 1. That felt dangerous, so I nixed some derived state from `this.state` ... which meant updating state using the result of the mutation. Do you like the change?

@caiosba (Also, I noticed some HTML injection in `<ReportImagePreview>` and instinctively changed it. TL;DR `.contentText` is safe; `.innerHTML` is dangerous. I got a bit carried away here, with Hooks. Did I go too far?)

@caiosba (And finally, I tried nixing Appium, as I was seeing tons of warnings in integration tests. Travis hums more happily now.)

@amoedoamorim The scariest bits you might care about: `<MoreLess>` logic was crashing during `mount()` in unit tests, so I rewrote it. And I was running into integration-test flakiness regarding `#login-or-register` (since its state was indeterminate). I wrote it as a `#register` button and a `#login` button to keep things safe.

caiosba:
Thanks, I took a look at the code on Friday, I'll look again today and most important, I'll make sure that all report designer functionality still works.

src/app/components/UploadImage.js
```diff
@@ -36,105 +37,117 @@ const StyledUploader = styled.div`
...
     if (file.size && unhumanizeSize(max_size) < file.size) {
-      if (this.props.onError) {
-        this.props.onError(file, this.props.intl.formatMessage(
-          messages.fileTooLarge,
-          { size: max_size },
-        ));
-      }
+      onError(file, <FormattedMessage
+        id="uploadImage.fileTooLarge"
+        defaultMessage="The file size should be less than {size}. Please try with a smaller file."
```
amoedoamorim, line 126: need to pass {size} value here?

infojunkie:
Please merge this to QA after this week's deployment

src/app/components/UploadImage.js
amoedoamorim: Other than this, :lgtm:! Just wait to merge after deployment as per karim's comment

PR #635, opened 2020-06-15
#8347 react intl cleanups

adamhooper:
Nix `isRtl`.

Start nixing `injectIntl()` and `defineMessages()`.

My goal is to upgrade React-intl and thus nix some warnings [#8179] ... and there's lots to do along the way.

_Big changes_ that flowed naturally while rewriting:

* In several places, switched to material-UI CSS (`makeStyles()` and `withStyles()`) instead of styled-components [#8249]
* Build a `<SocialIcons>` component. Include a Slack icon.
    * Update `react-icons` to v3, and switch to material-ui icons instead of `react-icons` wherever possible. This increases dev bundle size by a lot (because react-icons v3 doesn't allow importing icons by-path); but it doesn't include any unused icons in the production bundle.
* Nix `<Offset>` -- I don't think anybody knew what it did.
* `<TeamMembers>`: big rewrite. Nix `<RCTooltip>` and use `@material-ui/core/Tooltip` instead. Also, fix list-item JSX hierarchy to mirror material-ui docs [#8249]. Nix the `Relay.RootContainer` in `UserTooltip`, so there's just one query on the page that loads all tooltips at once. Upgrade to Relay Modern [#6211].
* `<DestinationProjects>`: big rewrite. Nix `react-select` and use `@material-ui/core/Autocomplete` instead, making a _much_ nicer dialog with less code. [#8249] Nix the `Relay.RootContainer` and upgrade to Relay Modern [#6211].
* `<TeamHeader>` -- material-ui-ize [#8249] ... though I couldn't figure out how to show this without literally editing the code to change an if-statement.
* `<AutoCompleteMediaItem>` -- fix HTTP-request races and errors using React Hooks, which makes this quite elegant. (If I understand correctly, the 2-second timeout before searching (even though we displayed "Loading...") was to avoid races? Well, now that races won't happen I lowered it to 1 second.) Nix `injectIntl()`, and use the material-ui default spot for the "Loading" and "Error" messages. [#8249] [#8347]

Spots for testing:

* `<TeamMembers>`: the view per workspace that lets you add and edit team members.
* The "Move" and "Add" buttons, both on the Search page ("bulk" buttons) and the Item page (not-"bulk" buttons) -- for `<DestinationProject>`.
* `<TeamHeader>` -- though I don't know how to show it.
* `<AutoCompleteMediaItem>` -- the "Create Related Media" dialog, under "Add Existing"


props.theme.dir
```diff
@@ -0,0 +1,46 @@
...
+~                                                                               
+~                                                                               
+~                                                                               
+~                                                                               
+~                                                                               
+~                                                                               
+~                                                                               
+~                                                                               
+[0;7m[38;2;147;161;161m[48;2;7;54;66mprops                                                         0,0-1          All[m[38;2;131;148;150m[48;2;0;43;54m
+                                                                                [H[34h[?25h
```
amoedoamorim, line 46: wow, what is this?

adamhooper: aaaaah no idea! Nixed on localhost, will push

adamhooper:
This pull request is obsolete. These changes are incorporated into other pull requests now.

PR #651, opened 2020-07-02
Feature/8347 react intl upgrade 5

adamhooper:
More changes to make upgrading `react-intl` smoother [#8347]

* Scrap `MediaUtil` for parts.
    * `<MediaTitle>`: new component, replaces `MediaUtil.title()` and declares itself as a Relay Modern fragment [#6211]
    * `<MediaTypeDisplayName>`: new component, replaces `MediaUtil.mediaTypeLabelFormattedMessage()` and `MediaUtil.mediaTypeLabel()`
    * `createdAt()`: now `helpers.parseStringUnixTimestamp()`. (Of all the callers of `createdAt()`, _none_ passed a ProjectMedia object. All of them forged fake objects to meet the calling convention.)
* Nix many `injectIntl()` and `defineMessage()` in:
    * `<DatetimeRespondTask>` (and nix unused error message)
    * `<DatetimeTaskResponse>` (and use a `<FormattedDate>` ... leading to a rewrite so https://mantis.meedan.com/view.php?id=8437 is now _only_ a problem on the server side; the new code comments describe how to adjust the client side once we let users select datetime+timezone.)
    * `<TimeBefore>` (and use `<FormattedRelative>` add a `dateTime` attribute to the `<time>`; use a `<FormattedDate>` to format the `title` attribute instead of `.toISOString()`)
    * `<MediaComponent>` (and de-indent, sowing chaos in the diff)
    * `<ReportDesignerForm>`
    * `<CreateProject>` (and add `propTypes`)
    * `<ProjectEdit>`
    * `<UserAssignments>` (and nix the `<div>` between the parent `<ul>` and its child `<li>`s)
    * `<JoinTeamComponent>`
    * `<SmoochBotTextEditor>`
    * `<TeamBots>` (and nix unused confirm-uninstall message)
    * `<TeamComponent>` (which didn't use `this.props.intl` at all)
    * `<TeamInfoEdit>`
    * `<Teams>`
    * `<AddTeamPage>`
    * `<SwitchTeamsComponent>`
    * ... nixing `injectIntl()` causes a change in unit tests: `mountWithIntlProvider(<(injectIntl(ComponentThatReturnsNull)) ... />).html()` used to erroneously return `''`; after nixing `injectIntl()`, `mountWithIntlProvider(<ComponentThatReturnsNull ... />`).html()` correctly returns `null`.
* `<BulkActions>`: fix React Relay warnings about missing props [#8179]; set buttons `disabled` when there's no selection
* `<MediaTags>`: put `key` on `<li>`, not its child [#8179]
* `pusher.js`: use `React.forwardRef()` to fix a warning about function components not accepting `ref` [#8179]
* `<CreateTeamCard>`, `<FindTeamCard>`: nix invalid+ignored prop, `titleStyle` [#8179]
* `UpdateProjectMediaMutation`: don't mutate Relay cache
* `spec/me_page.rb`: nix unused method


src/app/components/media/MediaTitle.js
```diff
@@ -0,0 +1,189 @@
...
+  children: PropTypes.func.isRequired,
+};
+
+function mediaTypeOrNull(projectMedia) {
+  switch (projectMedia.domain) {
+  case 'twitter.com': return MediaType.Twitter;
+  case 'facebook.com': return MediaType.Twitter;
+  case 'instagram.com': return MediaType.Twitter;
+  case 'youtube.com': return MediaType.Twitter;
+  case 'tiktok.com': return MediaType.Twitter;
```
caiosba, line 67: why all cases return `MediaType.Twitter`?

src/app/components/task/DatetimeRespondTask.js
```diff
@@ -228,8 +213,10 @@ class DatetimeRespondTask extends Component {
...
             onChange={this.handleChange.bind(this)}
-            okLabel={this.props.intl.formatMessage(messages.ok)}
-            cancelLabel={this.props.intl.formatMessage(messages.cancel)}
+            okLabel={<FormattedMessage id="datetimeRespondTask.ok" defaultMessage="OK" />}
+            cancelLabel={
+              <FormattedMessage id="datetimeRespondTask.cancel" defaultMessage="Cancel" />
+            }
             labelFunc={
               this.state.date ?
                 date => this.props.intl.formatDate(date) : null
```
caiosba, line 222: It was not possible to replace `intl.formatDate` by `<FormattedDate />` here?

src/app/components/media/MediaTitle.js
adamhooper: Aaaaaah! Great catch. Thank you! Will fix.

src/app/components/task/DatetimeRespondTask.js
adamhooper: I can't make a risk-vs-reward case for nixing `intl` here.

`@material-ui/pickers` is ... well ... a turd. That's why there's `this.state.date` _and_ `date` in this bit of code: `this.state.date` is _our_ state (which we need). `date` is the picker's state (even though it should be using a prop). We don't know when they differ.

`labelFunc` must return a `String`: a `<FormattedMessage>` isn't allowed. So we have two options:

A) Assume `this.state.date === date` always, and wrap the whole component in a `<FormattedMessage>`
B) Leave the `intl` call.

I see great risk in option A: we'd be trusting the code _we know to be wrong_ to behave the way we expect.

caiosba: Ah, gotcha... thanks!

src/app/components/media/MediaTitle.js
adamhooper: @caiosba Fixed. Also, I tested one of each type. And I added a new `MediaTitle.test.js` to test all cases. (It uncovered a bug -- woo-hoo!)

PR #655, opened 2020-07-15
Feature/7824 team custom statuses

amoedoamorim:


localization/react-intl/src/app/components/team/Statuses/DeleteStatusDialog.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "id": "deleteStatusDialog.statusInUseTitle",
+    "defaultMessage": "You need change the status of {items_count} items to delete this status"
```
caiosba, line 4: Missing word: "you need *to* change"

localization/react-intl/src/app/components/team/Statuses/DeleteStatusDialogBody.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "id": "deleteStatusDialogBody.statusInUseTitle",
+    "defaultMessage": "You need change the status of 76 items to delete this status"
```
caiosba, line 4: Missing word ("you need *to* change") and hard-coded value ("76") instead of placeholder.

src/app/components/layout/ConfirmProceedDialog.js
```diff
@@ -0,0 +1,63 @@
...
+          <FormattedMessage
+            id="confirmProceed.confirm"
+            defaultMessage="Confirm and proceed"
+          />
+        }
+      </Button>
+    </DialogActions>
+  </Dialog>
+);
+
```
caiosba, line 62: If we are going to keep this component (see comment about `<ConfirmDialog />`), please add `propTypes`

src/app/components/layout/ConfirmProceedDialog.js
```diff
@@ -0,0 +1,63 @@
...
+          <FormattedMessage
+            id="confirmProceed.learnMore"
+            defaultMessage="Learn more"
+          />
+        </Button>
+        : null
+      }
+      <Button className="confirm-proceed-dialog__cancel" onClick={onCancel}>
+        { cancelLabel ||
+          <FormattedMessage {...globalStrings.cancel} />
```
caiosba, line 42: If we are going to keep this component (see comment about `<ConfirmDialog />`), please define `PropTypes` with a default value.

src/app/components/layout/ConfirmProceedDialog.js
```diff
@@ -0,0 +1,63 @@
...
+        color="primary"
+        disabled={proceedDisabled}
+        onClick={onProceed}
+        variant="contained"
+      >
+        { proceedLabel ||
+          <FormattedMessage
+            id="confirmProceed.confirm"
+            defaultMessage="Confirm and proceed"
+          />
```
caiosba, line 56: If we are going to keep this component (see comment about `<ConfirmDialog />`), please define `PropTypes` with a default value.

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+import ConfirmProceedDialog from '../../layout/ConfirmProceedDialog';
+
+const DeleteStatusDialog = ({
+  deleteStatus,
+  onCancel,
+  onProceed,
+  open,
+  statuses,
+}) => {
+  const [moveToStatus, setMoveToStatus] = React.useState(0);
```
caiosba, line 16: Since status is not an integer, can we have a default value of `null` instead of zero?

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+          values={{ items_count: deleteStatus.items_count }}
+        />
+      }
+      body={
+        <React.Fragment>
+          <p>
+            <FormattedMessage
+              id="deleteStatusDialog.statusInUseMessage"
+              defaultMessage="There are {items_count} items with the status {statusName} that must be moved in order to delete this status, or you can go to each item and edit them individually."
+              values={{ items_count: deleteStatus.items_count, statusName: deleteStatus.label }}
```
caiosba, line 49: Can we have a single pattern for these placeholders? Either snake_case for everything or CamelCase for everything (I think the latter makes more sense for JavaScript).

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+            />
+          </p>
+          <p>
```
caiosba, line 52: I think we are trying to wrap those texts in `<Typography />` components for standardized styling sake, instead of `<p />` tags.

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+              values={{ published_count: deleteStatus.published_reports_count }}
+            />
+          </p>
+          <div>
+            <FormControl variant="outlined" fullWidth>
+              <Select
+                id="delete-status-dialog__select"
+                value={moveToStatus}
+                onChange={handleSelect}
+              >
```
caiosba, line 65: I think this should be a `<Menu />` instead of a `<Select />`.

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+                  ))
+                }
+              </Select>
+            </FormControl>
+          </div>
+        </React.Fragment>
+      }
+      onCancel={onCancel}
+      onProceed={handleSubmit}
+      proceedDisabled={moveToStatus === 0}
```
caiosba, line 91: I think this can be simply `{!moveToStatus}`

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+    />
+  );
+};
+
+DeleteStatusDialog.propTypes = {
+  deleteStatus: PropTypes.object,
+  onCancel: PropTypes.func.isRequired,
+  onProceed: PropTypes.func.isRequired,
+  open: PropTypes.bool.isRequired,
+  statuses: PropTypes.array.isRequired,
```
caiosba, line 107: Since the component expects each status to have `id` and `label`, would be good to change this `PropType` to be `arrayOf` `shape` with those two fields.

src/app/components/team/Statuses/EditStatusDialog.js
```diff
@@ -0,0 +1,187 @@
...
+  color: ${opaqueBlack38};
+`;
+const StyledHexValue = styled.span`
+  font: ${caption};
+  color: ${opaqueBlack87};
+`;
+
+const StyledSliderContainer = styled.div`
+  width: 80%;
+`;
```
caiosba, line 48: We agreed to use `makeStyles` instead of styled components from now on, including using `theme.spacing` instead of `units`.

caiosba: Also, if there is a good reason to use styled components instead of `makeStyles`, let's discuss!

src/app/components/team/Statuses/EditStatusDialog.js
```diff
@@ -0,0 +1,187 @@
...
+
+const EditStatusDialog = ({
+  defaultLanguage,
+  onDismiss,
+  onSubmit,
+  open,
```
caiosba, line 54: For all dialog-like components, not only this one... what do you think about hiding the `open` prop and instead of that just rendering / not rendering the component? Usually I do this when state doesn't need to be kept between showing/hiding.

src/app/components/team/Statuses/StatusListItem.js
```diff
@@ -0,0 +1,118 @@
...
+import IconMoreVert from '@material-ui/icons/MoreVert';
+import styled from 'styled-components';
+import globalStrings from '../../../globalStrings';
+import { subheading2 } from '../../../styles/js/shared';
+
+const StyledStatusLabel = styled.span`
+  color: ${props => props.color};
+  font: ${subheading2};
+  font-weight: 500;
+`;
```
caiosba, line 19: Please try to use `makeStyles` + Material UI components (like `<Typography />` with the relevant variation) instead of styled components.

src/app/components/team/Statuses/StatusListItem.js
```diff
@@ -0,0 +1,118 @@
...
+    handleClose();
+    if (onDelete) onDelete(status);
+  };
+  const handleEdit = () => {
+    handleClose();
+    if (onEdit) onEdit(status); // TODO default props to avoid checking?
+  };
+  const handleMakeDefault = () => {
+    handleClose();
+    if (onMakeDefault) onMakeDefault(status); // TODO default props to avoid checking?
```
caiosba, line 42: Since `onDelete`, `onEdit` and `onMakeDefault` are marked as required props, do we need those `if`'s?

src/app/components/team/Statuses/StatusesComponent.js
```diff
@@ -0,0 +1,309 @@
...
+    whiteSpace: 'nowrap',
+  },
+  helpIcon: {
+    color: checkBlue,
+  },
+}));
+
+const StyledBlurb = styled.div`
+  margin-top: ${units(4)};
+`;
```
caiosba, line 50: Can we replace this by `makeStyle`?

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+import { StyledStatusLabel } from './StatusListItem';
+
+const StyledTranslateStatusesContainer = styled.div`
+  margin: ${units(4)};
+`;
+
+const StyledColHeader = styled.div`
+  display: flex;
+  justify-content: space-between;
+`;
```
caiosba, line 22: Same about `makeStyles` x `styled-component`

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+        };
+        status.locales = locales;
+      }
+
+      return status;
+    });
+
+    setShowWarning(false);
+
+    if (onSubmit) {
```
caiosba, line 55: Since `onSubmit` is marked as `required`, do we need this `if`?

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+    newTranslations[id] = text;
+    setStranslations(newTranslations);
+  };
+
+  return (
+    <StyledTranslateStatusesContainer>
+      <Grid container spacing={2}>
+        <Grid item xs={6}>
+          <Typography variant="button">
+            {languagesList[defaultLanguage].nativeName}
```
caiosba, line 91: Please don't assume that the `defaultLanguage` is in `languagesList` - please fallback to the locale when there is no native name for that language. Make we could implement a helper for that.

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+          </Typography>
+        </Grid>
+        <Grid item xs={6}>
+          <StyledColHeader>
+            <Typography variant="button">
+              {languagesList[currentLanguage].nativeName}
```
caiosba, line 97: Please don't assume that the defaultLanguage is in languagesList - please fallback to the locale when there is no native name for that language. Make we could implement a helper for that.

src/app/components/team/Statuses/EditStatusDialog.js
```diff
@@ -0,0 +1,187 @@
...
+    </Dialog>
+  );
+};
+
+EditStatusDialog.propTypes = {
+  defaultLanguage: PropTypes.string.isRequired,
+  onDismiss: PropTypes.func.isRequired,
+  onSubmit: PropTypes.func.isRequired,
+  open: PropTypes.bool.isRequired,
+  status: PropTypes.object,
```
adamhooper, line 180: The interplay between this and state is unclear to me. It took me a lot of reading to figure out that we're treating this as an uncontrolled component.

Should we rename `status` to `defaultValue`? That would make it all clear.

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+  justify-content: space-between;
+`;
+
+const TranslateStatuses = ({
+  statuses,
+  defaultLanguage,
+  currentLanguage,
+  onSubmit,
+}) => {
+  const [translations, setStranslations] = React.useState({});
```
adamhooper, line 30: Typo. Should be `setTranslations`

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+
+    if (onSubmit) {
+      onSubmit(newStatusesArray);
+    }
+  };
+
+  const handleSave = () => {
+    let missingTranslation = false;
+
+    statuses.forEach((s) => {
```
adamhooper, line 63: Array.prototype has a function for this: `some()` is more efficient and expressive than `forEach()` here. `const missingTranslation = statuses.some((s) => ...)`

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+    if (missingTranslation) {
+      setShowWarning(missingTranslation);
+    } else {
+      handleSubmit();
+    }
+  };
+
+  const handleTextChange = (id, text) => {
+    const newTranslations = { ...translations };
+    newTranslations[id] = text;
```
adamhooper, line 82: I know it's a super-detailed nit, but ... one-liner? `setTranslations({ ...translations, [id]: text })`

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+        }
+        onCancel={handleDialogCancel}
+        onProceed={handleSubmit}
+      />
+    </StyledTranslateStatusesContainer>
+  );
+};
+
+TranslateStatuses.propTypes = {
+  statuses: PropTypes.array.isRequired,
```
adamhooper, line 158: `arrayOf(PropTypes.shape({ ... }).isRequired).isRequired`?

src/app/components/search/SearchResultsTable/StatusCell.js
```diff
@@ -34,12 +34,14 @@ export default function StatusCell({ projectMedia }) {
...
-        id: PropTypes.string.isRequired,
-        label: PropTypes.string.isRequired,
-      }).isRequired), // undefined during optimistic update
-    }).isRequired,
+    team: PropTypes.shape({
+      verification_statuses: PropTypes.shape({
+        statuses: PropTypes.arrayOf(PropTypes.shape({
+          id: PropTypes.string.isRequired,
+          label: PropTypes.string.isRequired,
+        }).isRequired), // undefined during optimistic update
```
adamhooper, line 42: This comment confuses me. Why is `verification_statuses` undefined during optimistic update? Can we fix that?

src/app/components/team/Statuses/TranslateStatuses.js
```diff
@@ -0,0 +1,164 @@
...
+        title={
+          <FormattedMessage
+            id="translateStatuses.missingTranslations"
+            defaultMessage="Missing translations"
+          />
+        }
+        body={
+          <FormattedMessage
+            id="translateStatuses.missingTranslationsBody"
+            defaultMessage="Some statuses are missing translation. If they need to be used in this language, they will be replaced by the same value in the default language, and may not be legible to users."
```
adamhooper, line 147: Agreement error in the first sentence. Also, does it run a bit long? How about:

> Some statuses are missing translations. Users may not be able to read untranslated statuses.

src/app/components/team/Statuses/StatusesComponent.js
```diff
@@ -0,0 +1,309 @@
...
+      onError: (error) => {
+        handleError(error);
+        setDeleteStatus(null);
+      },
+    });
+  };
+
+  const handleDelete = ({ status_id, fallback_status_id }) => {
+    submitDeleteStatus({
+      team_id: team.id,
```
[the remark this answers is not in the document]
caiosba: Yes, almost all mutations take a GraphQL `ID`.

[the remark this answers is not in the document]
caiosba: Agree should be consistent.

caiosba:
> The code is looking great. I pointed out some nits, grammar, variable-name stuff ... nothing structural.
> 
> However, @caiosba @amoedoamorim I do ring one architectural alarm bell: it looks like we're telling the _user_ that Statuses are GraphQL Objects. Users see verbs like "Add", "Edit" and "Delete". They see Items and Reports connected to Statuses. Yet these behaviors are all faked on the client side.
> 
> What happens if the user deletes a used status in `updateTeam()`? Modifies an ID? Submits two statuses with the same ID? Creates an Item with an invalid ID? And so on. What happens when _we_ want to change the Status schema? These events will occur in production sooner or later. GraphQL Objects present a standard set of patterns and expectations -- the patterns and expectations that match our users' understanding.
> 
> Is it our intent to make Statuses into GraphQL Objects? If so, what is the path from where we are now to GraphQL Objects?

As these setting structures get more complex, we may want to turn those settings into separate objects themselves. This happened on the past with other team settings like `TeamTask` and `TagText`... they were simple JSON settings when they didn't have much logic attached them and then at some point they turned into GraphQL objects with their own logic. But for now, team statuses are all still a JSON structure in team settings, as you could see... and we try to be consistent by having a JSON Schema on the backend that validates the format.

src/app/components/team/Statuses/EditStatusDialog.js
amoedoamorim: Sorry, I might have misunderstood. I thought we agreed either `makeStyles` or styled components were fine and we were just dropping inline styles.

amoedoamorim: A good reason to use them is they are more straightforward to use and read, imo

src/app/components/team/Statuses/EditStatusDialog.js
adamhooper: @caiosba The `open` prop is used to manage transitions. Stuff happens during open and close, and unmounting the component would hide all that.

adamhooper: (think "fade-in", "fade-out" animations.)

caiosba: Ahh that's a good reason, thanks

src/app/components/layout/ConfirmProceedDialog.js
```diff
@@ -0,0 +1,63 @@
...
+        }
+      </Button>
+    </DialogActions>
+  </Dialog>
+);
+
+export default ConfirmProceedDialog;
```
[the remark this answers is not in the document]
amoedoamorim: There are some differences:

`ConfirmDialog` requires the user to mark a checkbox in order to proceed. It keeps track of the checkbox state and blocks the submit button accordingly.

`ConfirmProceedDialog` is a more flexible type of dialog. It allows for the content to be anything. I made it to enable the design provided by Pierre. In this case the dialog keeps no state, it's just an arrow function component. In this usage the body contains a status selection component in order to indicate which status should be used in substitution to the one being deleted.

I wanted to keep each component very simple doing one kind of interaction each and not having to manage special cases or conditions. I think we can keep both and use each in the situation it's best suited to.

src/app/components/team/Statuses/DeleteStatusDialog.js
amoedoamorim: I don't understand this one

src/app/components/layout/ConfirmProceedDialog.js
```diff
@@ -0,0 +1,63 @@
...
+    open={open}
+  >
+    <DialogTitle>
+      {title}
+    </DialogTitle>
+    <DialogContent>
+      {body}
+    </DialogContent>
+    <DialogActions>
+      { onLearnMore ?
```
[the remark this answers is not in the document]
amoedoamorim: Alright, this was in the designs. Need to make @pierreconti aware of this.

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+
+  if (!deleteStatus || !open) return null;
+
+  return (
+    <ConfirmProceedDialog
+      open={open}
+      title={
+        <FormattedMessage
+          id="deleteStatusDialog.statusInUseTitle"
+          defaultMessage="You need change the status of {items_count} items to delete this status"
```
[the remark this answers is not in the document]
amoedoamorim: This was in the designs too, but I agree it's too long for a title. @pierreconti

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+          values={{ items_count: deleteStatus.items_count }}
+        />
+      }
+      body={
+        <React.Fragment>
+          <p>
+            <FormattedMessage
+              id="deleteStatusDialog.statusInUseMessage"
+              defaultMessage="There are {items_count} items with the status {statusName} that must be moved in order to delete this status, or you can go to each item and edit them individually."
```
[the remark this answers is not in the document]
amoedoamorim: This is the copy provided in the designs. @pierreconti

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -0,0 +1,114 @@
...
+              values={{ items_count: deleteStatus.items_count, statusName: deleteStatus.label }}
+            />
+          </p>
+          <p>
+            <FormattedMessage
+              id="deleteStatusDialog.itemsPublishedMessage"
+              defaultMessage="{published_count} of those items are currently published. Upon moving them to another status, the reports will be paused. Please review those items to re-publish them."
```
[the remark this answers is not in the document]
amoedoamorim: Will add this to the ticket for @pierreconti 's review too

src/app/components/team/Statuses/DeleteStatusDialog.js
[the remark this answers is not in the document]
pierreconti: Thanks for flagging @amoedoamorim - Will review asap with Kat. This should have happened earlier, my bad.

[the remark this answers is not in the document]
pierreconti: Here you go @amoedoamorim : `There are {items_count} items with the status {statusName} that must be changed to other statuses before delete this status. Alternatively, you can change each item statuses individually.`

src/app/components/team/Statuses/TranslateStatuses.js
pierreconti: Agree. Please use the copy suggested by @adamhooper. Apologize for not providing the copy for this modale @amoedoamorim. Please don't hesitate to reach out (Figma/Slack prob the best at the design stage) if scenarios are missing from Figma. .

src/app/components/team/Statuses/DeleteStatusDialog.js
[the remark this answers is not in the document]
pierreconti: I disagree on this one – it is indeed an alert. Users are alerted about something they (likely) do not expect. It's action oriented and still concise. The point of putting it in the title is that the title is likely the only few words that will be read.  'Delete status' does not raise the important issue caused by the deletion.

src/app/components/layout/ConfirmProceedDialog.js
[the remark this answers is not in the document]
pierreconti: I agree there should not be 'learn more' - I did take it off the object in Figma, but the one you saw it on was detached (my fault). I was not aware, though, that we were changing the modals style, as this is an app wide change that should be reflected on the roadmap. Not that i don't want to (we need to), but we would probably have had more conversations about it (like this, and other buttons). Perhaps I'm missing a piece? Thanks for raising both. Happy to chat more about modals.

src/app/components/layout/ConfirmProceedDialog.js
```diff
@@ -0,0 +1,63 @@
...
+        className="confirm-proceed-dialog__proceed"
+        color="primary"
+        disabled={proceedDisabled}
+        onClick={onProceed}
+        variant="contained"
+      >
+        { proceedLabel ||
+          <FormattedMessage
+            id="confirmProceed.confirm"
+            defaultMessage="Confirm and proceed"
```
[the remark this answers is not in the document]
pierreconti: This should be part of the modal conversation. Let's schedule that.

src/app/components/team/Statuses/StatusesComponent.js
```diff
@@ -0,0 +1,309 @@
...
+      },
+      onError: (error) => {
+        handleError(error);
+        setDialogOpen(false);
+      },
+    });
+  };
+
+  const handleAddOrEditStatus = (status) => {
+    const newStatuses = { ...team.verification_statuses };
```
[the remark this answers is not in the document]
amoedoamorim: In this case I'll keep as is. I'm splicing the copied array because I want to keep the edited status at the same index it previously was.

src/app/components/search/SearchResultsTable/StatusCell.js
amoedoamorim: This comment was already there. It seems it was added by you in PR #610. I'm not sure what is going on there.

src/app/components/team/Statuses/DeleteStatusDialog.js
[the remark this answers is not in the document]
amoedoamorim: Now it won't show the dialog and the status will be deleted right away if not in use.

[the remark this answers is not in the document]
amoedoamorim: Done. @infojunkie could you review the copy to ensure it looks good for l10n plz?

src/app/components/team/Statuses/DeleteStatusDialog.js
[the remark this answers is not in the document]
amoedoamorim: Done. @infojunkie could you review the copy to ensure it looks good for l10n plz?

src/app/components/search/SearchResultsTable/StatusCell.js
adamhooper: Hahaha! Well, that's embarrassing :). I guess it's not worth tracking down what I was thinking....

PR #659, opened 2020-07-17
<MoveProjectMediaToProjectAction> in Relay Modern [#6211]

adamhooper:
* Nix UpdateProjectMediaProjectMutation, and use an inline one
* Keep logic out of <MediaActionsBar>
* Be a Relay Fragment Container
* Never query `ProjectMedia.projects`. It turns out we only used it in CreateRelatedMedia, for the getCurrentProject() helper. In the future, we're going to nix that. Currently, mutations don't mutate this connection even when they should. I think it's easiest to ignore it and never use it.

I'm working from the leaves to the root, relay-modernizing and making
anything that uses Relay data into a fragment container. One button at a
time!


adamhooper:
Hiya @amoedoamorim,

This small pull request tries to solve stuff from the bottom up. I'm interested in experimenting more with the idea of moving mutations into leaf components -- in this example, the same file as the `<Button>`. What do you think of this approach? (And of the code?)

This mutation causes Relay Store cache inconsistency ... but that isn't a regression. https://mantis.meedan.com/view.php?id=8497 suggests the previous code also led to an inconsistent cache.

adamhooper:
(rebasing to fix Travis+imgur)

amoedoamorim:
> This small pull request tries to solve stuff from the bottom up. I'm interested in experimenting more with the idea of moving mutations into leaf components -- in this example, the same file as the `<Button>`. What do you think of this approach? (And of the code?)

I like this approach and the code looks good. Not sure if this same pattern should be replicated everywhere. In the statuses editing I'm working on, for example, I'm keeping the mutation code in the parent `StatusesComponent` because it is shared by different actions in sub components (add/edit/make default).

But I like the approach and support using it when it makes sense as a self-containing action.

> This mutation causes Relay Store cache inconsistency ... but that isn't a regression. https://mantis.meedan.com/view.php?id=8497 suggests the previous code also led to an inconsistent cache.

If it's not a regression, seems we can handle it in #8497

adamhooper:
> I like this approach and the code looks good. Not sure if this same pattern should be replicated everywhere. In the statuses editing I'm working on, for example, I'm keeping the mutation code in the parent StatusesComponent because it is shared by different actions in sub components (add/edit/make default).

I think the reason `StatusesComponent` actions all share the same mutation is that the API presents a single mutation to handle the four distinct user actions (add/edit/delete/make-default). Is that a good reason?

When we have four UX paths all invoking one API method, should we: A) build four client-side mutations; or B) build one client-side mutation that does four different things?

I think we have experienced great costs when we have tried to model UX paths on API mutations. For example, look how hard it is to maintain `optimisticProjectMedia()`. It's complex, and it hinders more than it helps when moving/removing/archiving/restoring items. We use it in all these UX paths because the "create" user action needs it.

My hypothesis: it would have been cheaper to code all the `UpdateProjectMediaMutation` and `BulkUpdateProjectMediaMutation` UX paths as unrelated mutations, even though that doesn't mirror the API.

There are pros and cons to both approaches. I think now's a good time for experimentation :). Let's see what happens!

caiosba:
> My hypothesis: it would have been cheaper to code all the `UpdateProjectMediaMutation` and 
> `BulkUpdateProjectMediaMutation` UX paths as unrelated mutations, even though that doesn't mirror the API.

I think it's much better. I recently extracted the bulk-archive and bulk-restore item operations as separate mutations:
* https://github.com/meedan/check-web/blob/develop/src/app/relay/mutations/BulkArchiveProjectMediaMutation.js
* https://github.com/meedan/check-web/blob/develop/src/app/relay/mutations/BulkRestoreProjectMediaMutation.js

Much easier to read, test and maintain... and the only `if`'s we have is to check if the item belongs to a list or not.

amoedoamorim:
> I think the reason StatusesComponent actions all share the same mutation is that the API presents a single mutation to handle the four distinct user actions (add/edit/delete/make-default). Is that a good reason?

The way I see it, in this case it makes sense to reuse the same mutation code on the client for add/edit/delete operations because they're ultimately the same operation, we're editing the `team.verification_statuses` field. The delete operation is another mutation, it takes another set of args and has other consequences then the first one. But I guess I'd split them if they were completely unrelated operations working on different fields even if the API offered one single mutation to change all of them.

adamhooper:
@caiosba Asking for you to take a look, too, as you've recently been fixing the mutation in question.

src/app/components/media/MoveProjectMediaToProjectAction.js
```diff
@@ -0,0 +1,217 @@
...
+  return commitMutation(Relay.Store, {
+    mutation: graphql`
+      mutation MoveProjectMediaToProjectActionMutation($input: UpdateProjectMediaProjectInput!) {
+        updateProjectMediaProject(input: $input) {
+          project_media_project {
+            id
+            project_id
+            project {
+              id
+            }
```
caiosba, line 26: This is redundant - we already have `project { id }` in line 37 below

PR #661, opened 2020-07-21
#7824 Query pm.project_ids; Bold status name in dialog

amoedoamorim:


src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -51,9 +51,9 @@ const DeleteStatusDialog = ({
       body={
         <React.Fragment>
           <p>
-            <FormattedMessage
+            <FormattedHTMLMessage
               id="deleteStatusDialog.statusInUseMessage"
-              defaultMessage="{itemsCount, plural, one {There is one item} other {There are # items}} with the status {statusLabel} that must be changed to other statuses before deleting this status."
+              defaultMessage="{itemsCount, plural, one {There is one item} other {There are # items}} with the status <b>{statusLabel}</b> that must be changed to other statuses before deleting this status."
               values={{ itemsCount: deleteStatus.items_count, statusLabel: deleteStatus.label }}
```
caiosba, line 57: I remember @adamhooper saying something about `FormattedHTMLMessage` on the latest `react-intl` version, but I don't remember... but I think it should be something like:
```
<FormattedMessage
  id="deleteStatusDialog.statusInUseMessage"
  defaultMessage="{itemsCount, plural, one {There is one item} other {There are # items}} with the status {statusLabel} that must be changed to other statuses before deleting this status."
  values={{ itemsCount: deleteStatus.items_count, statusLabel: <b>{deleteStatus.label}</b> }}
/>
```
In order to avoid it. Please confirm, Adam.

adamhooper: I confirm it.

... in other news, prefer `<strong>` to `<b>`. For decades, `<b>` was deprecated in favor of `<strong>`. It would appear the standards groups never managed to destroy `<b>`, so now they've come up with a _new_ element, `<b>`, which means "bring-attention-to" and has a very convoluted meaning (along the lines of "we don't know what `<b>` means").

src/app/components/media/ReportDesigner/index.js
```diff
@@ -22,6 +22,7 @@ const ReportDesignerContainer = Relay.createContainer(ReportDesignerComponent, {
         demand
         description
         last_status
+        project_ids
```
adamhooper, line 25: What does this do? :)

src/app/components/team/Statuses/DeleteStatusDialog.js
```diff
@@ -51,9 +51,9 @@ const DeleteStatusDialog = ({
       body={
         <React.Fragment>
           <p>
-            <FormattedMessage
+            <FormattedHTMLMessage
```
adamhooper, line 54: I think you can just use `<FormattedMessage>` (after making Caio's change below). Could you please confirm?

PR #670, opened 2020-07-30
Document and refactor GraphQL

infojunkie:
Companion PR to https://github.com/meedan/check-api/pull/611

- Ensure consistency in naming, API access, etc.
- Remove need to `JSON.parse` GraphQL fields where possible


src/app/components/search/SearchResultsTable/DemandCell.js
```diff
@@ -3,11 +3,11 @@ import PropTypes from 'prop-types';
 import NumberCell from './NumberCell';
 
 export default function DemandCell({ projectMedia }) {
-  const { demand } = projectMedia;
-  return <NumberCell value={demand} />;
+  const { requests_related_count } = projectMedia;
```
adamhooper, line 6: This file should be renamed to `RequestsRelatedCountCell.js`.

src/app/components/search/SearchResultsTable/LinkedItemsCountCell.js
```diff
@@ -3,11 +3,11 @@ import PropTypes from 'prop-types';
 import NumberCell from './NumberCell';
 
 export default function LinkedItemsCountCell({ projectMedia }) {
-  const linkedItemsCount = projectMedia.linked_items_count;
-  return <NumberCell value={linkedItemsCount} />;
+  const { related_count } = projectMedia;
```
adamhooper, line 6: This file should be renamed to `RelatedCountCell.js`.

Also ... if this is `RelatedCount`, should the other one `RequestsCount`? What's the different between a "request" and a "request related"?

src/app/relay/mutations/UpdateProjectMediaMutation.js
```diff
@@ -76,20 +76,21 @@ class UpdateProjectMediaMutation extends Relay.Mutation {
...
       });
       return {
         project_media: {
           id: this.props.media.id,
-          title: embed.title,
-          description: embed.description,
-          metadata: JSON.stringify(embed),
+          title: metadata.title,
+          description: metadata.description,
+          metadata: JSON.stringify(metadata),
```
adamhooper, line 93: Should this be JSON-stringified?

src/app/relay/mutations/UpdateReportDesignMutation.js
```diff
@@ -36,7 +35,7 @@ class UpdateReportDesignMutation extends Relay.Mutation {
     return {
       id: this.props.id,
       set_fields: JSON.stringify(dynamic.fields),
-      annotation_type: dynamic.annotation_type,
+      annotation_type: 'report_design',
```
adamhooper, line 38: While we're fixing things ... the next line, `annotated_id`, should not be Stringified.

src/app/components/media/Media.js
```diff
@@ -40,9 +40,11 @@ const MediaContainer = Relay.createContainer(MediaComponent, {
                 text
                 parsed_fragment
                 annotator {
-                  id
-                  name
-                  profile_image
+                  ... on User {
```
adamhooper, line 43: This is a dangerous pattern. If `annotator` is a `BotUser`, these fields will not exist. In Relay Classic any access will give `undefined`. In Relay Modern, any access will crash (white screen of death).

I searched the code and couldn't find where `annotator` is used. There's one destructuring bit in `MediaTimelineUtils.js` (which would crash, given a `BotUser`) and that's it.

infojunkie:
> Most of these changes look spectacular.
> 
> One change looks like an anti-pattern: changing the API from today's `dynamic_annotation_analysis` to tomorrow's`dynamic_annotation_analysis: annotation(annotation_type: "analysis")`.
> 
> GraphQL is strongly typed. These "dynamics" are weakly typed. This "dynamic" API idea invalidates Point 3 on the three selling points at https://graphql.org/.
> 
> Has any API user asked for this change?
>

We are trying to reduce the surface area of the API by removing `dynamic_annotation_XXXX`. These GraphQL types are actually automatically generated on the backend, and they are all synonymous to `annotation` with a different `annotation_type`. In addition, some API users have custom-made annotations that we don't want to expose as "official" annotation types.

> This is an anti-pattern because the correct connection name here is `analyses`. 
>

Good point, we can change the alias to `analyses`, and so on for different annotation types.

> Ask 20 GraphQL experts, and they'll all say so.
> 

GraphQL was released in 2015, and we've been using since 2016. Does that make us experts too?

src/app/components/media/Media.js
infojunkie: Yes, thanks, I think we should ensure `Annotation.annotator` always refers to ` User` (instead of being polymorphic), and that `BotUser` is a superset of `User`.

src/app/components/search/SearchResultsTable/LinkedItemsCountCell.js
infojunkie: we have 2 types of request counts: requests made on the item itself (`requests_count`), and requests made on the item + all its related items (`requests_related_count`)

PR #671, opened 2020-07-30
Feature/8418 content language

amoedoamorim:


caiosba:
> can we make Rails return `Team.languages`, an Array of Strings, and then obliterate `Team.get_languages`? @caiosba?

Yes, this is happening in a branch that is refactoring some things on the GraphQL layer.

src/app/components/team/Languages/AddLanguageAction.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  const handleSubmit = () => {
+    const onCompleted = () => {
+      setSelectedLanguage('');
+      setDialogOpen(false);
+    };
+    const onError = () => {};
+
+    submitAddLanguage({
```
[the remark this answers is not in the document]
amoedoamorim: I'm using the `isSaving`/`CircularProgress` pattern here. As with the statuses, we don't have individual permissions for languages so it could be a bit tricky to block menu actions on optimistic data.

src/app/components/team/Languages/LanguageListItem.js
```diff
@@ -0,0 +1,167 @@
...
+    });
+  };
+
+  const handleMenuMakeDefault = () => {
+    handleClose();
+    setDefaultDialogOpen(true);
+  };
+
+  const submitDefault = () => {
+    const onCompleted = () => {
```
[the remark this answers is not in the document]
amoedoamorim: I'm using the `isSaving`/`CircularProgress` pattern here. As with the statuses, we don't have individual permissions for languages so it could be a bit tricky to block menu actions on optimistic data.

src/app/components/layout/ConfirmDialog.js
```diff
@@ -88,21 +88,32 @@ class ConfirmDialog extends React.Component {
...
+  title: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element,
+  ]).isRequired,
+  message: PropTypes.object,
+  blurb: PropTypes.node,
   disabled: PropTypes.bool,
-  checkBoxLabel: PropTypes.object,
-  continueButtonLabel: PropTypes.object,
+  checkBoxLabel: PropTypes.oneOfType([
```
adamhooper, line 109: I feel this should be `PropTypes.node` (which captures both).

src/app/components/layout/ConfirmDialog.js
```diff
@@ -88,21 +88,32 @@ class ConfirmDialog extends React.Component {
...
+    PropTypes.string,
+    PropTypes.element,
+  ]),
+  continueButtonLabel: PropTypes.oneOfType([
```
adamhooper, line 113: `PropTypes.node`

src/app/components/team/Languages/AddLanguageAction.js
```diff
@@ -0,0 +1,167 @@
...
+  const [dialogOpen, setDialogOpen] = React.useState(false);
+  const [isSaving, setIsSaving] = React.useState(false);
+  const [value, setValue] = React.useState(null);
+  const setFlashMessage = React.useContext(FlashMessageSetterContext);
+
+  const languages = safelyParseJSON(team.get_languages) || [];
+  const options = Object.keys(LanguageRegistry)
+    .filter(code => !languages.includes(code))
+    .sort((a, b) => compareLanguages(null, a, b));
+  const getOptionLabel = code => `${languageLabel(code)} (${code})`;
```
adamhooper, line 64: "X (Y)" should be a `<FormattedMessage>`. Ref: https://www.mediawiki.org/wiki/Localisation#Symbols.2C_colons.2C_brackets.2C_etc._are_parts_of_messages

src/app/components/team/Languages/LanguageListItem.js
```diff
@@ -0,0 +1,281 @@
...
+  onFailure,
+}) {
+  commitMutation(Store, {
+    mutation: graphql`
+      mutation LanguageListItemSetDefaultLanguageMutation($input: UpdateTeamInput!) {
+        updateTeam(input: $input) {
+          team {
+            id
+            get_language
+            get_languages
```
adamhooper, line 35: I don't think this changes ... does it?

src/app/components/team/TeamComponent.js
```diff
@@ -183,6 +184,19 @@ class TeamComponent extends Component {
...
+                label={
+                  <FormattedMessage
+                    id="teamSettings.languages"
+                    defaultMessage="Languages"
+                  />
+                }
+                value="languages"
+              />
+              : null }
+            {UserUtil.myRole(this.getCurrentUser(), team.slug) === 'owner' ?
```
adamhooper, line 199: Should be `currentUserIsOwner`

src/app/components/team/Languages/LanguageListItem.js
amoedoamorim: what do you mean?

src/app/components/team/Languages/AddLanguageAction.js
amoedoamorim: Interesting! Looks like I'll need to use  `injectIntl` and `props.intl` on this one as the Autocomplete component expects a string and performs `.toLowerCase` on it

adamhooper: Good catch. I think you can call it `renderOption` instead. (The default `renderOption()` implementation calls `getOptionLabel()`. I don't see why there are two methods in the first place....)

PR #691, opened 2020-08-28
#8632 Fix error message display and a bit of cleanup

amoedoamorim:


src/app/components/UserPasswordChange.js
```diff
@@ -53,44 +34,48 @@ class UserPasswordChange extends Component {
...
-              </CardContent>
-            </Card>
-          }
-        </StyledPasswordChange>
-      </PageTitle>
+                    <ChangePasswordComponent
+                      type="reset-password"
+                      show_current_password={false}
+                      token={token}
+                      show_confirm={this.showConfirm.bind(this)}
```
caiosba, line 68: I know this is not a change in this PR, but why those props use snake case instead of camel case?

amoedoamorim: I noticed it, but I didn't want to touch it in this commit.

PR #704, opened 2020-09-16
Feature/8081 new columns

amoedoamorim:


src/app/components/search/SearchResults.js
```diff
@@ -317,6 +318,19 @@ class SearchResultsComponent extends React.PureComponent {
 
     const isProject = !!this.props.project;
 
+    const smoochBotInstalled = (
```
infojunkie, line 321: is this condition present elsewhere in the code? I suggest to make a helper `isBotInstalled(team, identifier)` out of that.

src/app/components/search/SearchResultsTable/LastSubmittedCell.js
```diff
@@ -2,10 +2,10 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import TimeCell from './TimeCell';
 
-export default function LastSeenCell({ projectMedia }) {
+export default function LastSubmittedCell({ projectMedia }) {
   return <TimeCell unixTimestampInS={parseFloat(projectMedia.last_seen)} />;
```
infojunkie, line 6: since we're renaming things, we would also want to rename the API field too, right?

amoedoamorim: Doesn't feel like a needed change imo but I defer to you and @caiosba. Would we change other fields and sorting keys as well?

```
SORT_MAPPING = {
    'recent_activity' => 'updated_at', 'recent_added' => 'created_at', 'demand' => 'demand',
    'related' => 'linked_items_count', 'last_seen' => 'last_seen', 'share_count' => 'share_count'
  }
```

src/app/components/search/SearchResults.js
amoedoamorim: done

src/app/components/search/SearchResultsTable/LastSubmittedCell.js
caiosba: Defer to Karim as the one taking care of the GraphQL API refactoring.

PR #706, opened 2020-09-22
#8706 remove team header clutter from settings page

amoedoamorim:


src/app/components/team/TeamComponent.js
```diff
@@ -273,8 +266,7 @@ class TeamComponent extends Component {
         <div className="team">
           <HeaderCard>
             <ContentColumn>
-              <Message message={this.state.message} />
```
caiosba, line 276: Are error and success messages still displayed for the settings pages after this line was removed? I guess "yes" because they are now using the `FlashMessage` component?

amoedoamorim: Yes. This state was unused actually, so this was just a simple cleanup.

caiosba (approved):

danielevalverde (approved):

PR #729, opened 2020-10-11
Updated dockerfile to Node 12.19.0

ducksecops:


caiosba:
The Travis file must updated too, as well as `package-lock.json` and possibly `package.json` in order to be sure that tests run against that version and that all packages are compatible with that version.

PR #747, opened 2020-10-27
Updated Social Network Colors in shared.js

3t8:
Updated social network colors and added reference links in shared.js
I also think that googleOrange should be called googleRed but I haven't seen reference where the color is used


3t8:
Alright, I removed it

PR #750, opened 2020-10-29
Feature/8776 move item history to dialog

amoedoamorim:


src/app/components/media/ItemHistoryDialog.js
```diff
@@ -0,0 +1,50 @@
...
+      <DialogContent>
+        <MediaLog
+          media={projectMedia}
+          team={team}
+        />
+      </DialogContent>
+    </Dialog>
+  );
+};
+
```
caiosba, line 49: Please add the `PropTypes`.

caiosba (changes requested):
Looks good to me - just one thing: please add `PropTypes` to the `ItemHistoryDialog` component.

src/app/components/media/ItemHistoryDialog.js
```diff
@@ -0,0 +1,58 @@
...
+ItemHistoryDialog.propTypes = {
+  open: PropTypes.bool.isRequired,
+  projectMedia: PropTypes.object.isRequired,
+  team: PropTypes.object.isRequired,
```
amoedoamorim, line 54: I was wondering, in cases where the prop is just received to be passed down, is it okay to now specify what's the shape of the prop? Then specify it where the shape is actually needed?

caiosba: Yes, that's what I do... I specify only as `object` and then at the final "destination" I specify the actual shape.

PR #759, opened 2020-11-10
Fix syntax for invoking bash in parallel test script.

sonoransun:
This presents as an error in the log:
```
./parallel-test.sh: line 1: !/bin/bash: No such file or directory
3938Starting tests...
```
but is unlikely to change behavior. Fixing for correctness.


sonoransun:
Unlikely to have an effect for anyone, however, some distributions use weird default shells like csh or tcsh. On these platforms, this fix will ensure the test script runs correctly.

PR #798, opened 2021-01-29
Script to track remaining relay classic files

amoedoamorim:


caiosba (changes requested):
Looks good to me! But I think we should also track the Modern stuff, so we can compare.

caiosba:
Looks good to me! But I think we should also track the Modern stuff, so we can compare.

PR #872, opened 2021-06-18
CHECK-665: Integrate new check-ui Metadata components

dariusk:
This PR introduces two major changes:

 - integration of the new `check-ui` shared metadata components for feature parity with the refactored `check-mark`
 - upgrading major underlying libraries: `react` from 16 to 17 (plus `react-dom` and `enzyme` as a side effect), and `eslint` from 4 to 7

## Check UI integration

Generally speaking, the new metadata components are integrated into `Task.js` by introducing (messy) branching logic, in order to stay in line with the style in which `Task.js` is currently implemented. _At some point in the near future, we will refactor `Task` to be significantly cleaner and remove functionality that is no longer used but that was out of the scope of this feature._

- Workflow changes
  - Compatibility with the new [yalc](https://github.com/wclr/yalc) based `check-ui` workflow
    - yalc is a tool that lets you publish to a local mock npm-style repository while developing a package. This means you can iterate on an external package without publishing to npm dozens of times in a day, and without messing around with symlink hacks.
    - babel updated to parse commonjs modules
    - @meedan dependencies are now included in webpack's file watch, so that when a dev is working on `check-ui`, changes published through yalc get reflected in the local dev build
    - in the docker-compose configuration in `check`, `yalc` is installed and mapped to the check-ui build from `check-web`'s ' `/root/.yalc` directory to `check-ui`'s `.yalc` directory (this happens in `check` but is worth mentioning here)
  - add webpack `file-loader` so we can load the images in `leaflet-js`
  - defining a publicPath in webpack for leaflet assets
- Feature changes
  - we now use `check-ui` components for all metadata editing
  - add `about` field to all top-level metadata queries
  - Added some extra code to expect certain data structures when using the new metadata code vs the original task code.
  - Modifying HTML layout in Task component so tests can pass
  - Updating the metadata tests for check-ui functionality
  - Updating various other integration tests to add brief delays

## Underlying library upgrade

React was upgraded to 17 in order to be able to use `check-ui`'s new UI components.


caiosba:
Looks amazing Darius, thanks! Please just be sure if the `sleep`s in the integration tests are really needed.

PR #907, opened 2021-09-06
Ticket CHECK-601: The text from a destroyed comment is available only…

danielafeitosa:
… on object_changes

This change was included after paper_trail upgrade to version 10 on CheckAPI.
For comments deleted after the changes the `content` is null and need to get the text from the changes


amoedoamorim (changes requested):
Looks good, Dani. I just think the outer curly brackets are unneeded and maybe we can go with `safelyParseJSON` instead of `JSON.parse`.

PR #922, opened 2021-10-04
CHECK-937: conditional annotation "save all" and "required" fields

dariusk:
Overall this is a clientside implementation of "save all" which gets around the fact that we don't have that functionality on the graphql DB. We now hide the save/edit/delete buttons on the metadata fields, but they still exist in the DOM. We iterate over each hidden save button and check html data-elements in order to see whether a field is empty and whether it is required. If a field is required and empty, we don't allow the saves to happen. Otherwise, we simulate a click on each save button, thus saving all the fields.

- add a function that evaluates all metadata fields in the form and finds the most recently updated time and user who updated it
- use html data-elements to indicate on each Save button whether the fields are required or empty
- store a local copy of the overall metadata fields and mutate that for local display, applying those changes to the DB when we save (conditional display of fields uses this local copy)

### DIFF
--- a/localization/react-intl/src/app/components/task/Tasks.json
+++ b/localization/react-intl/src/app/components/task/Tasks.json
@@ -10,5 +10,25 @@
   {
     "id": "tasks.goToSettings",
     "defaultMessage": "Go to settings"
+  },
+  {
+    "id": "metadata.couldNotSave",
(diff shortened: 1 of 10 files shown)


dariusk:
Integration test failed due to instagram/facebook outage, rest of tests passing, will merge.

PR #1025, opened 2022-02-17
CHECK-1302: Relative date range search

dariusk:
* refactor the `DateRangeFilter` from React class to function component
* change `last_seen` to `report_published_at` and relabel as "Report published"
* change `published_at` to `media_published_at` and relabel as "Media published"
* introduce a `buildValueRelative` function to the query building section which generates an object that can be passed to the graphql query for a relative range search, specifying `conditon` (always `"less_than"` for now, but could change in the future as we add more conditions), `period` (number of days/weeks/months/years, called `relativeQuantityValue` in the frontend code), and `period_type` (which is d/w/m/y but mapped using the `relativeRanges` object for clarity in the frontend code and represented by `relativeRangeValue`)
* change the query cleanup code in `SearchResultsComponent` to accept our new `report_published` and `media_published_at` fields as well as the new `less_than` condition
* tests

@caiosba : should this line still remain `last_seen`? I am not using it in the dropdown anymore but it seems like smoochbot is a special case that should remain? https://github.com/meedan/check-web/blob/feature/1302-date-range/src/app/components/search/SearchResults.js#L372


src/app/components/search/DateRangeFilter.js
```diff
@@ -68,127 +88,194 @@ function parseEndDateAsISOString(moment) {
...
+    relative: 'relative',
+    none: '',
+  };
+  const getInitialRangeType = () => {
+    if (value && value[getValueType()]?.condition) return rangeTypes.relative;
+    if (value && value[getValueType()]?.start_time) return rangeTypes.startEnd;
+    return rangeTypes.none;
+  };
+  const [rangeType, setRangeType] = React.useState(getInitialRangeType());
+  const relativeRanges = {
```
dariusk, line 115: I used this kind of "struct of constants" pattern to provide easy remapping in case the API changes

melsawy: @dariusk I can change back-end to accept `days`, `weeks`, `months` or `years` instead of `d`, `w`, `m` or `y` in case it's cleaner and easier in front-end what do you think?

caiosba:
@dariusk as per our conversation, we can remove the special case for the Smooch Bot.

src/app/components/search/DateRangeFilter.js
dariusk: @melsawy I think it's okay as it currently is. It's good practice on the front end to be able to map variables like this anyhow. APIs change, and so do frontend models for what things mean. Or maybe we want to change all our frontend variables to be, who knows, Hungarian notation in the future. This just lets us define the mapping between the frontend and the API however we like, and keep all the "magic strings" in one place.

PR #1038, opened 2022-02-24
CHECK-1199: Quick item add

dariusk:
This PR changes the way "Add Item" works in both the instance where a user "normally" adds an item, and in the case where a blank media is imported from Fetch and a user needs to add an item to it.

The way the new flow works is: the modal opens up, and the user enters an optional claim, along with either an attachment __or__ a text/url. When the user hits "submit", a new item is created exactly as it used to be by this feature, but with `set_claim_description` added to the payload if there is a claim. The item is set to whatever the default verification status is int he workplace, but then if the user has picked a different verification status on the form, a second GraphQL mutation kicks off on the success of the first, which updates the status to the user-selected one.

Change list:

- Add `verification_statuses` to the `team` query on `Home` component
- pass the `team` object down from `CreateMedia` -> `CreateMediaDialog` -> `CreateMediaInput`
- create a new state `status` on `CreateMediaInput` which stores a mock string like `'undetermined'` or `'verified'` to represent the current state of the status dropdown. these strings match the verification_status strings set on the server, the only one locally hard coded is the default `'undetermined'`, the rest is populated from the `verification_statuses` portion of the `team` query in `Home`
- update `MediaStatusCommon` component to take a `quickAdd` boolean prop that, when true, overrides the "lock" state of the status dropdown (this way the component doesn't have to know anything about the media item to be set)
- update `MediaStatusCommon` component to take a `currentStatus` string prop that acts as an alternative to the `media.last_status` for indicating the current state of the dropdown. this enables us to update the local state of the dropdown before making any graphql mutations (since the user will be changing these values in the UX before the item exists in the db)
- updating `react-dropzone` dependency so it supports `disabled` prop
- add a new `image+video+audio` type to the `UploadFile` component so it shows relevant information about those three mime-types combined
- update the `BlankMediaButton` (which shows when a blank media item is imported from fetch) to accept the `team` object and pass it on to `CreateRelatedMediaDialog` -> `CreateMediaInput` (since the latter component now requires `team` as a prop)
- claim text is now set at the same time as the item is created via the `set_claim_description` variable
- in `CreateMedia` we now have a cascade of mutations where a mutation is first done to create the item, and then on success, the item's status is updated to whatever the user picked in the `CreateMediaInput` dialog (I left in the option for creating a new status whole-cloth (`CreateStatusMutation`) just in case there was an edge case where no default status is defined in a workspace
- styles of `CreateMediaInput` updated to match the design
- clear button created to clear out the file dropzone
- `CreateProjectMediaMutation` now returns some extra fields needed by the second mutation for setting status, and also returns `claim_description` for cache update reasons (probably not needed but better safe than sorry)
- some unit tests, fix integration tests

### DIFF
--- a/localization/react-intl/src/app/components/UploadFile.json
+++ b/localization/react-intl/src/app/components/UploadFile.json
@@ -12,7 +12,7 @@
     "defaultMessage": "Drop an audio file here, or click to upload a file (max size: {audio_max_size}, allowed extensions: {audio_extensions})"
   },
   {
-    "id": "uploadFile.fileMessage",
+    "id": "uploadFile.imageVideoAudioMessage",
     "defaultMessage": "Drop a file here, or click to upload a file (max size: {file_max_size}, allowed extensions: {file_extensions})"
   },
   {
--- a/localization/react-intl/src/app/components/media/CreateMedia.json
+++ b/localization/react-intl/src/app/components/media/CreateMedia.json
@@ -9,6 +9,6 @@
   },
   {
     "id": "createMedia.addNewItem",
-    "defaultMessage": "Add new item"
+    "defaultMessage": "Add item"
   }
 ]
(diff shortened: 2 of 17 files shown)


dariusk:
Tests are failing due to flakes, passing locally, so merging.

PR #1051, opened 2022-03-08
CHECK-1117: Speed up next/prev media navigation

dariusk:
The problem was that we do a database query every time the user hits "next" or "previous" while viewing an individual item just to find the url we need to redirect to, then we redirect to it. This commit fixes the issue and makes navigating using the prev/next icons much faster.

* When a user navigates into a media item directly from a search view, a `mediaNavList` of `dbid` values for the (maximum) 50 locally rendered search items is passed down into the media item. This is now used to render the next/prev item indicators, which means we no longer have to run a query every time the user clicks a next/prev item indicator. The only time a query is now run is when we are transitioning out of a window of previously-calculated search items. So for example, if we are on item 50 in a list 1 to 50, and we hit "next", then a single query will fire off and pull down items 51 to 100, which are then passed as a new `mediaNavList`. If the user hits "next" from there, it will not fire a query. But if the user hits "previous" from there, it WILL fire off a query and re-fetch items 1 to 50. So basically: we only do the extra query when going out of bounds of our pre-fetched list.
* The `mediaNavList` and `count` (total search item count) values are passed into new items via the `browserHistory.push` function, which in addition to accepting a path, will also accept an object containing `pathname`, `search`, and an arbitrary `state`. The `state` contains the `mediaNavList` and `count` values, and is basically a way of passing props via `browserHistory`.
* In the case where a user enters a hyperlink directly to an item that came from a search (such as typing it in the url bar or being linked from outside the application), we default to the old "slow" behavior where we re-run the query and do a query for both the previous and next buttons. However, once a user hits previous or next, it will then pass the new query in and use the fast component.
* Also fixed is a nasty bug where `pageSize` was defined as `50` in `SearchResults.js` but as `20` in `urlHelpers`. I made `urlHelpers` the canonical value and `SearchResults` now imports it. This fixes a bug where rounding errors resulted in a user's place being lost if they clicked into an item while on a second page of a search list and then hit the "back" button, due to rounding math errors because of the conflicting `pageSize` values.

### DIFF
--- a/src/app/components/media/MediaPage.js
+++ b/src/app/components/media/MediaPage.js
@@ -36,6 +36,8 @@ export default function MediaPage({ route, routeParams, location }) {
       listId={listId}
       projectMediaId={projectMediaId}
       view={currentView}
+      mediaNavList={location?.state?.mediaNavList}
+      count={location?.state?.count}
     />
   );
 }
--- a/src/app/components/media/MediaPageLayout.js
+++ b/src/app/components/media/MediaPageLayout.js
@@ -21,7 +21,7 @@ const StyledTopBar = styled.div`
 `;
 
 export default function MediaPageLayout({
-  listUrl, buildSiblingUrl, listQuery, listIndex, projectId, projectMediaId, view,
+  listUrl, buildSiblingUrl, listQuery, listIndex, projectId, projectMediaId, view, mediaNavList, count,
 }) {
   return (
     <div>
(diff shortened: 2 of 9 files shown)


caiosba:
Looks great! Fantastic improvement :)

dariusk:
All the integration tests pass locally, and Daniele told me we are having problems in the Travis CI environment so I'm going to go ahead and merge.

PR #1076, opened 2022-03-24
Feature/1651 friendly error messages

dariusk:
Work in progress


src/app/CheckError.js
```diff
@@ -10,6 +13,71 @@ const CheckError = {
     MISSING_OBJECT: 8,
     DUPLICATED: 9,
     LOGIN_2FA_REQUIRED: 10,
+    CONFLICT: 11,
```
dariusk, line 16: @caiosba  Can you look at these strings and tell me if they make sense? I did my best looking at the api source code, but since of these constants aren't even used by the API in the first place.

PR #1087, opened 2022-04-01
CHECK-1658: Allow removing rule from tag

amoedoamorim:


caiosba (approved):
Alex, please test for both the regular Rules UI and the rules component on the tags editing page.

PR #1106, opened 2022-04-19
Keep track of number of in-progress mutation requests

dariusk:
On the network layer, we modify a global `inFlightMutationRequests` variable that lives on `window` and can be accessed from any function. This was the most reasonable solution I could come up with for a small amount of code, even though it doesn't hook into the React lifecycle. The `HomeComponent` now calls `setRouteLeaveHook` with a function that tests to see if the number of in-progress mutation requests is greater than 0, and prompts the user to see if they really want to navigate away.

In the Jira ticket, the wording was to do this whenever _queries_ are still processing, but it doesn't make sense for non-mutation queries. A user should be allowed to navigate away from a page mid-READ, but prompted mid-WRITE.

This also removes an extraneous `withRouter` no longer needed in `Tasks.js` after Alex refactored it to use the `NavigateAwayDialog`. I am not using that new component here because this is a special case where I want the `setRouteLeaveHook` logic to be different and not coupled to the state change of a particular component. If I did that to `HomeComponent`, we would end up constantly rerendering the entire application, so it doesn't make sense here.

Fixes CHECK-1643.

### DIFF
--- a/src/app/CheckNetworkLayer.js
+++ b/src/app/CheckNetworkLayer.js
@@ -56,6 +56,8 @@ class CheckNetworkLayer extends Relay.DefaultNetworkLayer {
     const { setFlashMessage, ...otherOptions } = options;
     super(path, otherOptions);
     this.setFlashMessage = setFlashMessage || (() => null);
+    this.inFlightMutationRequests = 0;
+    window.inFlightMutationRequests = this.inFlightMutationRequests;
   }
 
   _parseQueryResult(result) {
@@ -152,6 +154,8 @@ class CheckNetworkLayer extends Relay.DefaultNetworkLayer {
   }
 
   _sendMutation(request) {
+    this.inFlightMutationRequests += 1;
+    window.inFlightMutationRequests = this.inFlightMutationRequests;
     const _interopRequireDefault = obj => obj && obj.__esModule ? obj : { default: obj };
 
     let init;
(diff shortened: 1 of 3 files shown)


dariusk:
@caiosba and @amoedoamorim, I need your sense here of whether it makes sense to use a global variable on the `window` object here to get this done, instead of a react state-type-thing. I tried for 1.5 days to do this the "react" way but ended up in all sorts of re-rendering loop madness. The code as-is in this PR does actually catch all unfinished mutations on an attempt to nav away in-app.

amoedoamorim:
Yeah, I think this should work. The state management at the top and re-rendering the whole app sounds like hell and I can't think of any reasons for not going with this solution!

caiosba (approved):
I think it's fine to keep it in `window`. Two additional main reasons:

1) Thread-safety... it's still single-threaded as far as I know, so we shouldn't run into race condition situations.

2) If we keep it in a `Home` state or in a shared context, I'm afraid of the performance impact of cascading those changes down to every nested component.

caiosba:
I think it's fine to keep it in window. Two additional main reasons:

Thread-safety... it's still single-threaded as far as I know, so we shouldn't run into race condition situations.

If we keep it in a Home state or in a shared context, I'm afraid of the performance impact of cascading those changes down to every nested component.

amoedoamorim (approved):

PR #1107, opened 2022-04-20
Redirect to not found when trends not available

amoedoamorim:
This commit changes Trends.js to get the `get_trends_enabled` field on Team and validate it in order to avoid attempting to load the trends view from teams where it's not available.
If trends is not available for that team, accessing the `/team-slug/trends` url should lead to the `/check/not-found` page.

Fixes CHECK-1752


dariusk (approved):
Looks good to me!

caiosba:
:lgtm:

PR #1111, opened 2022-04-22
Collect user data and integrate error page with Intercom

amoedoamorim:
This commit puts an ErrorBoundary around each Route's parent component.
This way each ErrorBoundary is under the IntlProvider and has Intercom in scope so we integrate and prefill support request messages when crashes happen.
ErrorBoundary component now gets a `component` prop to specify which component it is wrapping and this will be sent with the errbit report.
It is also collecting user name, dbid and email to send with the report, to help gain more insight/context on the crashes.

Note I: The @airbrake-js/browser calls to the Errbit server seem to ignore the extra data added as context, params and session, unless the error is a string or we cast a new instance from the raised error object. To be investigated. See github issue: https://github.com/airbrake/airbrake-js/issues/1193

Note II: MainErrorBoundary has been renamed to ErrorBoundary

Fixes: CHECK-1744


src/app/components/media/MediaComponent.js
```diff
@@ -294,6 +294,9 @@ class MediaComponent extends Component {
     media.quote = media.media.quote;
     media.embed_path = media.media.embed_path;
 
+    // console.log(this.props.trololo.meajude.xocxoc); // eslint-disable-line no-console
+    console.log(JSON.parse('bli')); // eslint-disable-line no-console
```
dariusk, line 298: Stray console log!

src/app/components/project/ProjectGroup.js
```diff
@@ -4,103 +4,106 @@ import { QueryRenderer, graphql } from 'react-relay/compat';
...
-          team {
+  <ErrorBoundary component="ProjectGroup">
+    <QueryRenderer
+      environment={Relay.Store}
+      query={graphql`
+        query ProjectGroupQuery($id: ID!) {
+          project_group(id: $id) {
             id
-            slug
-            permissions
```
dariusk, line 24: I see you moved `slug` and `permissions` into a subquery here, just making sure that was as intended. I don't see any code changes made in response to this move, which is why I ask, but maybe I'm missing one.

src/app/helpers.js
```diff
@@ -64,8 +64,12 @@ function getStatusStyle(status, property) {
  * Truncate a string and append ellipsis.
  */
 function truncateLength(str, length = 70) {
-  const dots = str.length > length ? '...' : '';
-  return `${str.substring(0, length)}${dots}`;
+  if (typeof str === 'string') {
```
dariusk, line 67: Ah nice catch, wouldn't want this attempting to `substr` an `Array`.

dariusk (commented):
Other than my comments above (specifically that console log), looks good!

src/app/components/project/ProjectGroup.js
amoedoamorim: I didn't. They're under `team` all along. I guess it's the crazy diff visualization that may suggest it, but thanks for pointing!

src/app/components/media/MediaComponent.js
amoedoamorim: Thanks! Fixing this! Not only a console.log but one with a runtime error in it! :sweat_smile:

src/app/components/error/ErrorBoundary.js
```diff
@@ -23,10 +31,24 @@ class MainErrorBoundary extends React.Component {
...
       this.airbrake.notify({
-        error,
+        // For some reason we need to reinstantiate the error or pass the error as string.
+        // Not doing this causes the additional info (contex, params, session) to be empty
+        error: new Error(error),
+        context: { component },
         params: { info: errorInfo },
+        session: { name, dbid, email },
+      }).then((response) => {
+        if (Intercom) {
```
caiosba, line 46: Didn't you need to declare "global Intercom" on top? The Linter didn't complain?

caiosba (approved):

PR #1112, opened 2022-04-26
Update default sort on suggested matches page, mock relay unit tests

dariusk:
This commit changes the default sort on the `/suggested-matches/` route to "Submitted" (`recent_added`).

Added a unit test and picked up and fixed an additional minor bug in the `propTypes` for the `Search` component.

Note on testing: by running the `render` component of a `ReactRelayQueryRenderer` and passing in the appropriate `error` and `props` in a object as an argument to `render`, we can mock the relay renderer and do things like inspect derived properties on the rendered component.

```js
    const renderedMockSearch = wrapper.find('ReactRelayQueryRenderer').props().render(
      {
        error: false,
        props: {
          team,
        },
      },
    );
```

Fixes CHECK-1782.

### DIFF
--- a/src/app/components/search/Search.js
+++ b/src/app/components/search/Search.js
@@ -134,7 +134,7 @@ Search.propTypes = {
   title: PropTypes.node.isRequired,
   icon: PropTypes.node,
   hideFields: PropTypes.arrayOf(PropTypes.string.isRequired), // or undefined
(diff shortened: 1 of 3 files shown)


dariusk:
Merging since this failed on a test that has since been fixed.

PR #1151, opened 2022-06-06
Clear the entire CloudFlare cache to avoid negative cache race condition

sonoransun:
During the last deployment we encountered a scenario where the `check-web` service had an old and new task instance running. This in turn led to a race condition where the new root document referenced a vendor bundle file that did not exist on the old task instance. The resulting HTML file is then cached by CloudFlare, causing errors when viewing the site.

This change clears the entire CloudFlare cache during deployment, which will eliminate the race condition observed during the last deployment. Performance impact is expected to be minimal.

In the future, if we wish to be more fine grained with cache control, the Enterprise plan with CloudFlare supports clearing cache by prefix.


dariusk (approved):
Generally looks good! Will this clear both QA and Live? __Edit: answer is yes, from our video call.__

caiosba:
Looks good! This means that we're clearing cache for live as well when a QA deployment happens, right? It will be like that until we're able to clear cache for only prefixed URLs?

PR #1153, opened 2022-06-06
Add conditional to render confirmation modal and add unit tests

danielevalverde:
Do not render second confirmation dialog when editing status if smooch bot is not installed on the team and add unit tests

Reference: Ticket CHECK-1908


amoedoamorim (approved):
Show de buela!

PR #1154, opened 2022-06-06
Add conditional to render 'Report' tab and 'Report' Page and add unit tests

danielevalverde:
Do not render Report tab or Report Page if smooch bot is not installed on the team and add unit tests

Reference: CHECK-1910

### DIFF
--- a/src/app/components/team/TeamComponent.js
+++ b/src/app/components/team/TeamComponent.js
@@ -358,6 +358,9 @@ export default createFragmentContainer(TeamComponent, {
       alegre_bot: team_bot_installation(bot_identifier: "alegre") {
         id
       }
+      smooch_bot: team_bot_installation(bot_identifier: "smooch") {
+        id
+      }
     }
   `,
 });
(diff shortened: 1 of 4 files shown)


src/app/components/team/TeamComponent.js
```diff
@@ -219,7 +219,7 @@ class TeamComponent extends Component {
                 value="statuses"
               />
               : null }
-            {isAdminOrEditor ?
+            {isAdminOrEditor && Boolean(team.smooch_bot) ?
```
amoedoamorim, line 222: This requires `smooch_bot` to be requested in the team fragment below

amoedoamorim, line 222: This requires `smooch_bot` to be requested in the team fragment below

src/app/components/team/TeamReport/TeamReportComponent.js
```diff
@@ -104,6 +105,11 @@ const TeamReportComponent = ({ team, setFlashMessage }) => {
     });
   };
 
+  if (!team.smooch_bot) {
```
amoedoamorim, line 108: This also requires smooch_bot to be requested in the team fragment. This time the query is on the parent component (in `TeamReport/index.js`)

amoedoamorim (changes requested):
Code looks good, but it needs to request `smooch_bot` in the correct places, otherwise it won't work!

PR #1164, opened 2022-06-13
Remove video annotation timeline

dariusk:
* delete all `MediaTimeline*` components and tests
 * remove `onTimelineCommentOpen`, `showVideoAnnotation`, and `onVideoAnnoToggle` props from the component tree
 * update to `check-ui@0.1.51`, which removes the `Timeline` component from the library
 * delete all video annotation integration tests
 * remove video annotation icon from static assets


dariusk:
Integration and unit tests all pass, so merging to develop. This was pretty straightforward to do, it just took a lot of tracking down and making sure I wasn't removing any code that other components depend on, and that I wasn't leaving in any vestigial code.

PR #1173, opened 2022-06-21
Fix claim attribute styling

dariusk:
Turns out we just needed to remove a custom CSS override to make this display correctly. The default `caption` style is working! Also added some `description` fields to localized text in related files.


dariusk:
Build is failing on an intentional failure from unrelated errbit work, passing otherwise so merging.

PR #1177, opened 2022-06-27
Render 'Claim missing modal' when claim has empty description

danielevalverde:
or blank spaces description and add unit tests

Reference: CHECK-1953

### DIFF
--- a/src/app/components/media/MediaFactCheck.js
+++ b/src/app/components/media/MediaFactCheck.js
@@ -37,7 +37,7 @@ const MediaFactCheck = ({ projectMedia }) => {
   const readOnly = projectMedia.is_secondary || projectMedia.suggested_main_item;
 
   const handleGoToReport = () => {
-    if (!claimDescription) {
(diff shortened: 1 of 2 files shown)


src/app/components/media/MediaFactCheck.js
```diff
@@ -37,7 +37,7 @@ const MediaFactCheck = ({ projectMedia }) => {
   const readOnly = projectMedia.is_secondary || projectMedia.suggested_main_item;
 
   const handleGoToReport = () => {
-    if (!claimDescription) {
+    if (!claimDescription || claimDescription.description.length === 0 || claimDescription.description.trim().length === 0) {
```
amoedoamorim, line 40: No need to test for `claimDescription.description.length === 0` as `claimDescription.description.trim().length === 0` already covers that!

amoedoamorim (changes requested):
Looks great, Dani! You can just simplify that one line and it's good to go! :partyparroticecream:

PR #1180, opened 2022-07-06
Do not crash when open Select Project Dialog

danielevalverde:
Do not crash when open select project dialog in a workspace that has privacy enabled on default folder if the user has no permission to see that project and add unit test

Reference: CHECK-2069

### DIFF
--- a/src/app/components/media/SelectProjectDialog.js
+++ b/src/app/components/media/SelectProjectDialog.js
@@ -56,17 +56,18 @@ function SelectProjectDialog({
 
   // Add "empty folders" to empty collections, so they appear in the drop-down as well
   projectGroups.forEach((pg) => {
-    if (!projects.find(p => p.project_group_id === pg.dbid)) {
+    if (projects.length > 0 && !projects.find(p => p.project_group_id === pg.dbid)) {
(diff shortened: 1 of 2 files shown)


src/app/components/media/SelectProjectDialog.js
```diff
@@ -56,17 +56,18 @@ function SelectProjectDialog({
...
 
   // Get a default folder
-  const defaultFolder = projects.filter(({ dbid }) => dbid === team.default_folder.dbid)[0];
+  // The value returned can be undefined if workspace default folder has privacy settings on
+  const defaultFolder = team.default_folder ? projects.filter(({ dbid }) => dbid === team.default_folder.dbid)[0] : null;
 
   // Lastly, sort options by title and exclude some options and defaultFolder to add it to the top of the list
   const filteredProjects = projects
-    .filter(({ dbid }) => !excludeProjectDbids.includes(dbid) && dbid !== defaultFolder.dbid)
+    .filter(({ dbid }) => team.default_folder && !excludeProjectDbids.includes(dbid) &&  dbid !== team.default_folder.dbid)
```
amoedoamorim, line 70: If `team.default_folder` is falsy, then `filteredProjects` will always return empty. Instead please do the safe navigation in the last bit: `dbid !== team.default_folder?.dbid`

PR #1181, opened 2022-07-07
Adding a pull request template

caiosba:
Adding a pull request template that tries to capture our best practices for frontend development and also that makes code review easier.


dariusk:
Maybe add something to the checklist like "if I add a new external dependency, I include a rationale for doing so and an estimate of the change in bundle size"

dariusk:
One thing we could add to my above suggestion is a link to [Bundlephobia](https://bundlephobia.com/), which is a browser tool that estimates bundle size of npm dependencies.

PR #1186, opened 2022-07-14
Fix TypeError: Cannot read properties of undefined (reading 'edges')

danielevalverde:
## Description

The PR Fix the "TypeError: Cannot read properties of undefined (reading 'edges')" 
References: Ticket CHECK-2034

## Type of change

## How has this been tested?
Added unit tests for the component to check that the fix works, I reverted the fixes and checked that the tests failed

The PR Fix the "TypeError: Cannot read properties of undefined (reading 'edges')"  by adding verifications before trying to iterate over objects on MediaActionsBar Component

## Checklist


src/app/components/media/MediaActionsBar.js
```diff
@@ -250,17 +250,21 @@ class MediaActionsBarComponent extends Component {
...
 
     const options = [];
-    media.team.team_users.edges.forEach((teamUser) => {
-      if (teamUser.node.status === 'member' && !teamUser.node.user.is_bot) {
-        const { user } = teamUser.node;
-        options.push({ label: user.name, value: user.dbid.toString() });
-      }
-    });
+    if (media.team.team_users?.edges) {
+      media.team.team_users.edges.forEach((teamUser) => {
```
amoedoamorim, line 254: This will work, but here's a tip: probably no need to wrap this in an IF block, you can simply use `media.team.team_users?.edges.forEach` and it should be enough!

amoedoamorim (approved):

PR #1191, opened 2022-07-21
Add disabled relay/unused-fields rule to files

dariusk:
## Description

Per discussion on the engineering team, I'm adding the eslint relay plugin and having it flag an error whenever graphql fields are not used _in the same file_ as the query/fragment.

This means that the dozens of places where we have a big query on a parent component and then pass the result down to a child in a different file _will_ trigger a linter error. Since we can't fix this all at once, I'm taking the same path we did with the `description` eslint rule for react-intl and selectively disabling it in all files. As we do work on these files in the course of normal Check development, we will refactor the queries to correctly follow a pattern where data fragments of minimal size are pulled in only when needed by the code.

I also took the time to remove a few unused fields in a few files. I also added some `description` fields in a few places and removed the corresponding `eslint-disable @calm/react-intl/missing-attribute` comment in the header.

References: CHECK-2142

## Type of change

Refactor

## How has this been tested?

N/A, this is putting in eslint rules for future use.

## Checklist

- [x] I have made needed changes to the README
- [x] If I added a new external dependency, I included a rationale for doing so and an estimate of the change in bundle size (e.g., checked in https://bundlephobia.com/)

### DIFF
--- a/src/app/components/InviteNewAccount.js
+++ b/src/app/components/InviteNewAccount.js
@@ -277,13 +277,11 @@ const InviteNewAccount = ({ teamSlug }) => (
           dbid
           name
           email
-          login
           accepted_terms
           team_user(team_slug: $teamSlug) {
             team {
               dbid
               name
-              slug
             }
             invited_by {
               name
--- a/src/app/components/trends/TrendsItem.js
+++ b/src/app/components/trends/TrendsItem.js
@@ -1,3 +1,4 @@
+/* eslint-disable @calm/react-intl/missing-attribute, relay/unused-fields */
 import React from 'react';
 import PropTypes from 'prop-types';
 import { QueryRenderer, graphql } from 'react-relay/compat';
@@ -40,7 +41,6 @@ const TrendsItem = ({ routeParams, location }) => {
                       team {
                         name
                         dbid
-                        country
                       }
                     }
                   }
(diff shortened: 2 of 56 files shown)


src/app/components/InviteNewAccount.js
```diff
@@ -277,13 +277,11 @@ const InviteNewAccount = ({ teamSlug }) => (
           dbid
           name
           email
-          login
```
dariusk, line 280: These were simple fixes because the component that uses the data is currently within the same file; so I just removed the fields as they are obviously never used.

src/app/components/trends/TrendsItem.js
```diff
@@ -40,17 +41,13 @@ const TrendsItem = ({ routeParams, location }) => {
                       team {
                         name
                         dbid
-                        country
```
dariusk, line 43: This file was one that I was testing how I might refactor. I ultimately didn't do the refactor here, but I did temporarily move the `TrendsItemComponent` into this file and determine which fields could be safely removed. We still need to move the fragment into `TrendsItemComponent` to really "fix" this but ultimately this at least removes some truly unneeded fields.

dariusk:
Note: not doing a squash-and-merge on this because I want the atomic commit for future reference in documentation for refactoring.

PR #1194, opened 2022-07-22
Change Alegre port to 3100 to avoid conflict on Mac Monterey

hartsick:
Updates configuration and tests to account for change in Alegre port.

For more detail see commit messages and https://github.com/meedan/check/pull/69


dariusk:
Looks good to me but I want @sonoransun to take a look since the docker-compose files are touched and it might affect prod in some way that I don't understand.

huslage (approved):
LGTM

PR #1203, opened 2022-08-04
Fix message when moving item from spam list

danielevalverde:
## Description
Fix message when moving item from spam list
Reference:Ticket CHECK-2199

## Type of change

## How has this been tested?

- Mark an item as spam
- Find item on spam list and using the click 'Not spam'
Check that message says "Items moved from Trash"

## Checklist

### DIFF
--- a/src/app/components/media/BulkActions.js
+++ b/src/app/components/media/BulkActions.js
@@ -133,14 +133,28 @@ class BulkActions extends React.Component {
         title: projectTitle,
         dbid: projectId,
       } = this.state.dstProj ? this.state.dstProj : { title: null, dbid: null };
-      const message = (
+      const message = this.props.page === 'trash' ? (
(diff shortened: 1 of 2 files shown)


src/app/components/media/BulkActions.js
```diff
@@ -133,14 +133,28 @@ class BulkActions extends React.Component {
...
+            ),
+          }}
+        />
+      ) : (
+        <FormattedMessage
+          id="bulkActions.movedFromSpamSuccessfully"
+          defaultMessage="Items moved from Spam to '{toProject}'"
+          description="Banner displayed after items are moved successfully"
+          values={{
+            toProject: (
```
melsawy, line 156: `toProject` is same for two cases so I think it's better to define `const toProject` and use it.

PR #1236, opened 2022-09-12
Replace similarity strings to match glossary

danielevalverde:
## Description

Replaced the strings below:
On the "folder" page: 
Similar media → Matched media
Suggested matches → Suggested media
Show matched media

On the "item" page: 
More media → Matched media
Detach media (in menu of single secondary media) → Un-match media

Reference: CHECK-2221


amoedoamorim (changes requested):
Good job, Dani. I'll request, though, that message id's remain unchanged. Let's change copy but keep the id's please. Changing an id essentially deletes the string from transifex and generates a new one, while, keeping the id is just a copy change to the previous string. This keeps history on Transifex and helps translation. Thank you! :hugging_face:

PR #1263, opened 2022-10-07
Feature/2057 matched suggested media

dariusk:
This commit changes the behavior of the "Matched media" button in the similarity panel to scroll down to the media list instead of loading a new page. This commit also attempts a strict implementation of a [Check Design System (CDS)](https://www.figma.com/file/rnSPSHDgFncxjXsZQuEVKd/Check-Design-System) component for the first time -- there is a new directory at `components/cds` that is organized to match the Figma CDS document organization.

Components in the CDS directory are meant to be strictly view components, with business logic and data passed into them as props. FragmentContainer components will still live in the usual places in the code but should import CDS components to render data retrieved from the API wherever possible.

References: CHECK-2057

## Type of change

## How has this been tested?

Tested in dev, unit tests written, demo'ed for design.

- File naming and organization! Do we like `src/app/components/cds/buttons-checkboxes-chips/CounterButton.js`?
- Take a look at `src/app/styles/js/shared.js` to see the theming for typography in action and that we are overloading the `h2` typography variant for what we're calling "Title" in Check Design System. Material-UI v5 supports renaming variants so we can have them match better but for now I have to find variants we aren't currently using (most of the header ones), mapping them, and putting a note in both the code and also as a comment in the CDS figma.

## Checklist

- [X] I have made needed changes to the README
- [X] I have removed the /* eslint-disable @calm/react-intl/missing-attribute */ from any files I've worked on and added a `description` prop to all `<FormattedMessage />` components that were missing it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ before_script:
 jobs:
   include:
     - stage: tests
-      if: branch = develop OR branch = master
+      # if: branch = develop OR branch = master
       name: integration-and-unit-tests
       script:
       - docker-compose exec
@@ -72,7 +72,7 @@ jobs:
         web npm run tests
     - stage: tests
       name: similarity-tests
-      if: branch = develop OR branch = master
+      # if: branch = develop OR branch = master
       script:
       - docker-compose exec
         -e IMGUR_CLIENT_ID=$IMGUR_CLIENT_ID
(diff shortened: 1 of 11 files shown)


amoedoamorim (approved):
This looks awesome! :partyparrot:

PR #1325, opened 2023-01-13
Paginated, bulk UI for accept/reject/spam/trash suggested media

dariusk:
First of all, this PR contains a refactor of `MediaSuggestionsComponent` such that it is now paginating the data for suggested similar media. The component is now the child of `PaginatedMediaSuggestions`, which is the child of `MediaSuggestions`. Both of the two latter components are graphql wrappers and all application logic is in `MediaSuggestionsComponent`.

The overall reasoning and methodology is explained in detail on our Confluence wiki in the "How to paginate in Relay 1.7" article in the Engineering Cookbook. That page also explains some Relay 1.7 compiler quirks that required me to duplicate some queries instead of referencing them. The self-contained commit for pagination is a17a004.

I wrote some unit tests for `MediaSuggestionsComponent` around pagination. 

There is a `SuggestedMediaDialogComponent` which is the popup you get when you focus on a single suggested item in the suggestions list by clicking on its card. In the full design, this is its own paginated component that lets the user navigate item-by-item through the entire list -- it should be its own ticket, and I'll write that one up.

### One big file (future refactors)

In general, `MediaSuggestionsComponent` is way too long, at 1069 lines of code. There are a few obvious places to address this.

There are 6 dialog boxes, separately defined, that comprise about 100 lines of code. Componentizing these dialogs is necessary but we really should do them such that they are CDS components -- these would be the "Custom vertical" and "Import data to workspace" components under "Menus, lists and Dialog" in Figma. Ultimately this needs to be turned into a CDS dialog box component and it should be six lines of code here like

```
<ConfirmationDialog title={foo} confirmButtonText={bar} onConfirm={doThing} />
<ConfirmationDialog title={baz} confirmButtonText={biz} onConfirm={doThing2} />
<ConfirmationDialog folderList={folders} title={title} confirmButtonText={text}  onConfirm={doThingWithFolders} />
```

...or something like that. Per CDS we have a `Dialog` component defined with six variants, but maybe there is some work to be done around conceptual grouping.. for example some of these are pure display components and some of these have forms embedded.

Anyway, instead of componentizing this early, I figured I would copy/paste code here and make a bookmark to come back when it's time to work on developing the confirmation dialog component.

Another place we should refactor `MediaSuggestionsComponent` is moving the graphql into some other file. We have a few inline query declarations that really ought to be moved out, and would cut probably 200 lines from the file.

### DIFF
--- a/src/app/components/media/MediaComponentRightPanel.js
+++ b/src/app/components/media/MediaComponentRightPanel.js
@@ -46,7 +46,7 @@ const MediaComponentRightPanel = ({ projectMedia, setShowTab, showTab }) => {
             label={
               <FormattedMessage
                 id="mediaComponent.suggestedMedia"
-                defaultMessage="Suggested media"
+                defaultMessage="Suggestions"
                 description="Label for the 'Suggested media' tab, to show a list of media items that are suggested as similar to the one the user is viewing"
               />
             }
--- a/src/app/components/media/MediaComponentRightPanel.js
+++ b/src/app/components/media/MediaComponentRightPanel.js
@@ -90,7 +90,7 @@ const MediaComponentRightPanel = ({ projectMedia, setShowTab, showTab }) => {
       { /* Set maxHeight to screen height - (media bar + tabs) */ }
       <Box maxHeight="calc(100vh - 112px)" style={{ overflowY: 'auto' }}>
         { showTab === 'requests' ? <MediaRequests media={projectMedia} all={!projectMedia.is_confirmed_similar_to_another_item} /> : null }
-        { showTab === 'suggestedMedia' ? <MediaSuggestions projectMedia={projectMedia} /> : null }
+        { showTab === 'suggestedMedia' ? <MediaSuggestions dbid={projectMedia.dbid} teamDbid={projectMedia.team?.dbid} /> : null }
         { showTab === 'metadata' ? <MediaTasks media={projectMedia} fieldset="metadata" /> : null }
         { showTab === 'source' ? <MediaSource projectMedia={projectMedia} /> : null }
         { showTab === 'notes' ? <MediaComments media={projectMedia} /> : null }
--- a/src/app/components/media/Similarity/MediaSuggestions.js
+++ b/src/app/components/media/Similarity/MediaSuggestions.js
@@ -20,11 +20,22 @@ const mediaSuggestionsQuery = graphql`
       demand
       ...PaginatedMediaSuggestions_root
       team {
+        check_search_trash { id, number_of_results }
+        check_search_spam { id, number_of_results }
+        public_team { id, trash_count, spam_count }
         slug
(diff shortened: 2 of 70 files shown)


src/app/components/media/Similarity/MediaSuggestions.js
```diff
@@ -3,77 +3,60 @@ import React from 'react';
...
+        medias_count
+        search_id
+      }
+    }
+  }
+`;
+
+const MediaSuggestions = ({ dbid, teamDbid }) => {
+  const ids = `${dbid},0,${teamDbid}`; // Project ID doesn't matter
+  const pageSize = 4;
```
dariusk, line 46: Page size is hard-coded here. Maybe it should go somewhere else?

src/app/components/media/Similarity/MediaSuggestionsComponent.js
```diff
@@ -415,49 +669,284 @@ const MediaSuggestionsComponent = ({
...
-            <Typography variant="body2">
-              <FormattedMessage
-                id="mediaSuggestionsComponent.question"
-                defaultMessage="Are the medias below a good match for the claim and medias on the left?"
-                description="Subtitle for similarity matching widget"
-              />
-            </Typography>
             <Box display="flex" alignItems="center">
               <>
+                <SuggestedMediaDialogComponent
```
dariusk, line 786: This div contains six different dialog box components. Ultimately this needs to be turned into a CDS dialog box component and it should be six lines of code here like

```
<ConfirmationDialog title={foo} confirmButtonText={bar} />
<ConfirmationDialog title={baz} confirmButtonText={biz} />
```

Instead of componentizing this early, I figured I would copy/paste code here and make a bookmark to come back when it's time to work on developing the confirmation dialog component.

src/app/components/media/MediaComponentRightPanel.js
```diff
@@ -90,7 +90,7 @@ const MediaComponentRightPanel = ({ projectMedia, setShowTab, showTab }) => {
       { /* Set maxHeight to screen height - (media bar + tabs) */ }
       <Box maxHeight="calc(100vh - 112px)" style={{ overflowY: 'auto' }}>
         { showTab === 'requests' ? <MediaRequests media={projectMedia} all={!projectMedia.is_confirmed_similar_to_another_item} /> : null }
-        { showTab === 'suggestedMedia' ? <MediaSuggestions projectMedia={projectMedia} /> : null }
+        { showTab === 'suggestedMedia' ? <MediaSuggestions dbid={projectMedia.dbid} teamDbid={projectMedia.team?.dbid} /> : null }
```
dariusk, line 93: We no longer pass the entire `projectMedia` object into `MediaSuggestions` -- all it requires is the `dbid` and the team `dbid` and then `MediaSuggestions` and `PaginatedMediaSuggestions` plug those in as variables to all their own graphql queries.

PR #1336, opened 2023-02-01
Typography to match figma design system

brianfleming:
## Description

The Check Design System will be going through some improvements/refinements, one will be to have a better collection of typography. The first step was to include all of the options that come standard with mui. This PR:
- Adds all of the [typography settings that are in Figma](https://www.figma.com/file/rnSPSHDgFncxjXsZQuEVKd/Check-Design-System?node-id=3106%3A36342&t=A6erSmK7CLgwy7UT-4), which will be used going forward so there is a 1-1 relationship with the codebase.
- Updates references to the few uses of the typography to a better choice with the current system. For instance, the "old" `h1` was visually more appropriate to be `h6`
- In the future if the typography is adjusted (most likely will happen) in figma, then I will update the references in the codebase again.
- I also just wanted to get a first PR out to be one of the cool kids at Meedan! 

## Type of change

## How has this been tested?

Since there were only a few changes, I manually inspected the changes

## Checklist

### DIFF
--- a/src/app/components/cds/buttons-checkboxes-chips/CounterButton.js
+++ b/src/app/components/cds/buttons-checkboxes-chips/CounterButton.js
@@ -45,7 +45,7 @@ const CounterButton = ({ label, count, onClick }) => {
         {label}
       </Typography>
       <br />
-      <Typography variant="h2">
+      <Typography variant="subtitle2">
         {count}
       </Typography>
     </Button>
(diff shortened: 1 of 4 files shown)


dariusk (approved):
This looks good! Do the changes in `shared.js`, specifically the removal of the comments, mean that we are changing Figma to match the material-ui v4 typography variant naming scheme (since we can't change our names until v5)?

brianfleming:
@dariusk yes that is correct. I believe I got it correct matching v4. For reference, here is the updated area of figma that contains all the typography settings we will be using going forward (caveat: like all things I'm sure it will be continually refined):

https://www.figma.com/file/rnSPSHDgFncxjXsZQuEVKd/Check-Design-System?node-id=3106%3A36342&t=m6NiRR1OJxTFF0mw-4

amoedoamorim:
Arrived a bit too late to this party but this looks great, Brian!

PR #1346, opened 2023-02-23
Disable ripple effect across app

brianfleming:
## Description

This PRs goal is to disable the ripple effect in material ui theme globally. See the reference issue for more info, but this is a basic step towards a more bespoke visual feel to Check

References: [CV2-2799](https://meedan.atlassian.net/browse/CV2-2799)

## Type of change

## How has this been tested?

Manual testing looking for any regressions. There should not be since this is an theme option.

- I believe this is the way to turn off all ripples, even if the code change targets "button" according to the interwebs this will impact all other components such as check boxes and during visual review this seems to be the case

### BEFORE
https://user-images.githubusercontent.com/418001/220990884-3f7ad691-aed1-4e44-ad37-dc0a7ee77146.mov

### AFTER
https://user-images.githubusercontent.com/418001/220990931-519fe16f-78f8-4bf3-b3d2-6099ad7a06a2.mov

## Checklist

[CV2-2733]: https://meedan.atlassian.net/browse/CV2-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CV2-2799]: https://meedan.atlassian.net/browse/CV2-2799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/styles/js/shared.js
+++ b/src/app/styles/js/shared.js
@@ -232,6 +232,11 @@ export const MuiTheme = {
       textTransform: 'uppercase',
     },
   },
+  props: {
+    MuiButtonBase: {
+      disableRipple: true,
+    },
+  },
   overrides: { // Override of all material UI components. Information at https://material-ui.com/api/{component}
     MuiTableCell: {
       stickyHeader: {
--- a/src/web/index/index.js
+++ b/src/web/index/index.js
@@ -56,7 +56,6 @@ const pusherContextValue = getPusherContextValueForClientSessionId(clientSession
 const dir = rtlDetect.isRtlLang(locale) ? 'rtl' : 'ltr';
 const styledComponentsTheme = { dir };
 const muiTheme = createMuiTheme({ direction: dir, ...MuiTheme });
-
 // JSS and StylesProvider and <Helmet><body> are to make material-ui
 // support right-to-left (e.g., Arabic).
 // See https://material-ui.com/guides/right-to-left/
--- a/src/web/index/index.js
+++ b/src/web/index/index.js
@@ -55,17 +55,7 @@ const pusherContextValue = getPusherContextValueForClientSessionId(clientSession
 
 const dir = rtlDetect.isRtlLang(locale) ? 'rtl' : 'ltr';
 const styledComponentsTheme = { dir };
-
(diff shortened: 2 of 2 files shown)


dariusk:
This looks good in terms of the change itself but I think you should make it in a different file -- you'll notice it imports an `MuiTheme`:

`import { MuiTheme } from '../../app/styles/js/shared';`

If we look at that file we'll see there's an `overrides` section that overrides default mui themes:

https://github.com/meedan/check-web/blob/22cfd88ad6be5449f32332ce264c0961392a1283/src/app/styles/js/shared.js#L235-L243

I suspect that if you put your customization in there, it should apply the same as it does here but this way it's in our customization file.

brianfleming:
@dariusk I tried that initially, but nothing happened. Let me give it another shot, perhaps I got some syntax incorrect the first time. If its not that then it is just some idiosyncrasy of material which also wouldn't be a surprise to me.

dariusk:
@brianfleming If you try again and can't get it to work, give me an example of a button where there is still ripple and and the code you changed, and I'll try and trace down why it isn't propagating.

brianfleming:
@dariusk ok. I moved the change to the shared file at least. I tried adding it to the overrides section, but no effect, so I added it as a props on the main theme object which has the correct effect. I added in some screen recordings to demonstrate in the main PR description. 
Thoughts? I think this satisfies what you are looking to accomplish?

dariusk:
Yeah, this does the trick! Weird that overrides seem to not work (or at least not in this case), I will file an issue to look into that.

dariusk (approved):

brianfleming:
> Yeah, this does the trick! Weird that overrides seem to not work (or at least not in this case), I will file an issue to look into that.

yeah definitely strange. I tried adding it as a new override on a new object `MuiButtonBase` as well as adding it to the existing `MuiButton` override and still no joy. 🤷

PR #1391, opened 2023-03-23
Heap in Dev

brianfleming:
## Description

@amoedoamorim I wanted to get your thoughts on this PR.
As I was exploring, I noticed the font call to google fonts uses a weight we don't use, so while i was going to remove it, I also noticed that we load Heap and uptime monitoring always. I thought there was no need to do this in any other env other than prod. So what do you think of this extremely minor optimization?

Additionally I thought the same about intercom, but it wasnt quite as straightforward on how to disable that for local environments.

## Type of change

## How has this been tested?

- Etc.

## Checklist

### DIFF
--- a/src/web/views/index.js
+++ b/src/web/views/index.js
@@ -1,3 +1,5 @@
+const NODE_ENV = process.env.NODE_ENV || 'production';
+
 const capitalize = (str) => {
   if (!str) {
     return '';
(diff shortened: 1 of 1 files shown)


brianfleming:
> Hi @brianfleming ! The usual way to not load them locally was simply not having those ids in the local or other config file. This way we can easily enable/disable them for testing purposes. I think it doesn't hurt to check for the environment for uptime and heap, but I'd keep Intercom as is because we do interact with it via code to pre-fill a message to support when the UI crashes and we might want to easily debug that whenever.

Oh I got you. I didnt notice that the server config was ignored. Thanks! closing

PR #1407, opened 2023-04-06
Add styelint

brianfleming:
## Description

Adding stylelint and custom rules to our new css build process

References: [CV2-2877](https://meedan.atlassian.net/browse/CV2-2877)

## Type of change

## How has this been tested?

Manual testing

### Build notes
- It is currently set to only lint css files that have changed
- for css module files (like the new text field component) it will only lint css files that are actually used. So in this branch, if you change something in the new text field css component that would trigger the linter, you will see no errors because its not used anywhere, however, if you add it to a file, you will see the lint errors
- I tried to follow the same convention as we use with eslint, so it "behaves" in a similar way

### Base rules
- The base set of rules that is used (then some overrides) is this config:
[stylelint-config-standard](https://raw.githubusercontent.com/stylelint/stylelint-config-standard/49652ebf38d6231b0402787a10def004e986ca08/index.js)

### Rules
- you can see a [full list of rules and definitions here](https://stylelint.io/user-guide/rules)
- The overridden rules are opinionated based on my experiences, but really all open to discussions, just wanted to start somewhere

### Not in this PR
- not sure how to get this into the CI pipeline if there is anything special we need to do

### Misc
- I linted all out existing files so everything passes. Failures should only happen when future developers touch a `.css` file
- I "ignored" the tooltip file because its not ours, so I just let it be what it is

## Checklist

[CV2-2877]: https://meedan.atlassian.net/browse/CV2-2877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caiosba (approved):
Code looks good to me. We can adapt the rules as we go.

dariusk:
Late to the party but yeah, I like this!

PR #1408, opened 2023-04-06
Icon proposal

brianfleming:
## Description

This is another opinionated proposal PR.

### Motivation
- Inspired by the epic to [refactor some front end items](https://meedan.atlassian.net/browse/CV2-2876)
- we would like to move away from material design and have a most bespoke design sense.
- we currently use material icons which is a great source for icons, but it is another dependency to maintain from a style perspective, just like the rest of the material ui
- we do have some custom icons, but they were done via js, this PR replaces that with plain svg. This opens up the ability to more easily add icons that are not material and custom to meedan
- this will give us the control to add whatever icons we want and not have to worry about them changing when updating the material icons

## Type of change

## How has this been tested?

manual testing to visually compare where icons have been replaced with existing qa environment

- This uses the [@svgr/webpack package](https://github.com/gregberge/svgr/tree/main) to put enable svg as react components
- the package ^ is versioned to work webpack 4 (what we currently use)

### Design role
- This will introduce a new responsibility for the design team: preflighting svg icons that are not currently in the code base.
- Currently when a developer needs to insert an icon, they just need to know the material design name, instead they now will need to reference the correct svg file in the codebase.
- The design team will take on the responsibility of delivering the necessary svg icons that are missing in a timely manner to the engineering team

### Future
- as we work we can slowly replace material icons with ones in the codebase. This PR demonstrates how to deal with various situations as a stress test and guide.

## Checklist


amoedoamorim (approved):
:lgtm:

PR #1424, opened 2023-05-02
CV2-3079: add lock_version for similarity settings

melsawy:
## Description
This feature handle conflict on similarity settings and should handle this situation 

- User A opens the similarity settings for a workspace
- User B opens the same page
- User B changes setting X value
- User B save the changes
- User A changes setting Y value
=> Setting X value will be reverted, because User A had the old value loaded in their form
Same as CV2-2482 ticket

References: CV2-3079

## Type of change

## How has this been tested?

- Open similarity settings page in two different tabs 
- Go to first tab and update settings then save
Result: settings updated
- Go to second tab and update settings
Result: should go an error because you load old values 

- Etc.

## Checklist


dariusk (approved):
Thanks for adding the error handler, I'm glad to see more verbose and useful errors here.

PR #1436, opened 2023-05-18
Fixes to newsletter header upload error states

dariusk:
The upload component now displays two upload-related server-side errors: when a file is too big and when it is the wrong type.

For example:

image

This commit also fixes a minor display bug in some versions of Chrome - other versions unaffected by the change.

References: CV2-2873

## Type of change

## How has this been tested?

Verified fixed locally by dev.

There is a `FIXME` in line 192 of `NewsletterComponent.js` where we are not localizing the string that the API returns when an uploaded file is too big. It was too variable for me to figure out how to turn into a localized string.

## Checklist

### DIFF
--- a/localization/react-intl/src/app/components/cds/inputs/Upload.json
+++ b/localization/react-intl/src/app/components/cds/inputs/Upload.json
@@ -16,7 +16,12 @@
   },
   {
     "id": "upload.dragMessage",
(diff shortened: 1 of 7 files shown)


caiosba:
@dariusk , regarding your comment, the backend should return it localized (once a translation exists).

caiosba (approved):

PR #1440, opened 2023-05-30
Disable newsletter settings save/schedule if no file

dariusk:
When there is no file attached and we have a video/audio/image header, we disable saving and scheduling of the newsletter.

For reviewers: does the conditional code in the `useEffect` statements make sense? I commented but perhaps I could be clearer. I just didn't want to be overly verbose either.


caiosba (approved):

caiosba (approved):
The `useEffect` logic is clear to me.

amoedoamorim (approved):

PR #1444, opened 2023-05-31
Some fixes for the link management and newsletter interfaces.

caiosba:
## Description

* UTM code field should not be required
* UTM code field should be editable only while the newsletter is paused
* Changed the copy link management warning box
* When characters left are negative in the introduction and articles field, the helper text should be red - the entire field should be in error state
* RSS: Auto add “https://” if it’s missing from the field when clicking the load button
* Header default should be link preview, it’s empty

## Type of change

## How has this been tested?

Just manually and visually.

## Checklist

### DIFF
--- a/src/app/components/team/Newsletter/NewsletterComponent.js
+++ b/src/app/components/team/Newsletter/NewsletterComponent.js
@@ -64,7 +64,7 @@ const NewsletterComponent = ({
   const [introductionText, setIntroductionText] = React.useState(introduction || '');
   const [articleNum, setArticleNum] = React.useState(number_of_articles || 0);
   const [articles, setArticles] = React.useState([first_article || '', second_article || '', third_article || '']);
-  const [headerType, setHeaderType] = React.useState(header_type || '');
+  const [headerType, setHeaderType] = React.useState(header_type || 'link_preview');
   const fileNameFromUrl = new RegExp(/[^/\\&?]+\.\w{3,4}(?=([?&].*$|$))/);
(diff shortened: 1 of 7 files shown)


src/app/components/team/Newsletter/NewsletterComponent.js
```diff
@@ -429,7 +429,7 @@ const NewsletterComponent = ({
             handleFileChange={handleFileChange}
             setFile={setFile}
             setFileName={setFileName}
-            availableHeaderTypes={['image', 'none'] || team.available_newsletter_header_types || []}
+            availableHeaderTypes={team.available_newsletter_header_types || []}
```
dariusk, line 432: Oops, thanks for getting rid of my debug code!

src/app/components/team/Newsletter/NewsletterRssFeed.js
```diff
@@ -77,7 +77,13 @@ const NewsletterRssFeed = ({
                     placeholder={placeholder}
                     disabled={disabled}
                     className={styles['rss-feed-url-field']}
-                    onChange={(e) => { setLocalRssFeedUrl(e.target.value); }}
+                    onChange={(e) => {
+                      let { value } = e.target;
+                      if (!/^https?:\/\//.test(value)) {
+                        value = `https://${value}`;
```
dariusk, line 83: I can't confirm from my phone but will this prepend https to an invalid url too? If yes, maybe check clientside whether appending https:// creates a valid URL, and if not, then don't actually commit the prepend

caiosba: It's almost like a prefix... the backend will return an error if it's not valid.

PR #1448, opened 2023-06-01
Validate field length in newsletter settings

dariusk:
This commit adds an `onErrorTooLong` function prop to `LimitedTextArea` -- a function that is called, and passed a binary error state as its first and only parameter. This allows controlling forms to set state variables etc as needed when a `LimitedTextArea`'s error state changes due to going over or under the character limit.

Both `NewsletterScheduler` (for introduction) and `NewsletterStatic` (for article field length) use this new hook to communicate to the form validation.


amoedoamorim (approved):

caiosba (approved):
Tried locally and works great!

PR #1451, opened 2023-06-01
Specifying en-US locale for string conversion

dariusk:
This is how the string is stored in the database -- by default the browser uses browser locale, which in a MM/DD place will produce a different outcome from DD/MM places.

Fixes CV2-3168


amoedoamorim (approved):

caiosba (approved):
Tried locally and works perfectly!

PR #1465, opened 2023-06-13
Display error message below RSS input field when RSS feed URL is invalid

caiosba:
## Description

We've been already applying the error state to the RSS feed URL input field when the URL is invalid, but it's more clear to also display an error message.

Fixes: CV2-3175.

## Type of change

## How has this been tested?

* Go to the newsletter settings page
* Turn on "RSS feed"
* Put an RSS feed URL that is not valid
* You should see an error message below the RSS feed URL input field

## Checklist


dariusk (approved):
Looks good! This is what displays after I hit "load"
image

PR #1467, opened 2023-06-13
Clear errors on newsletter header file upload component when file is cleared

caiosba:
## Description

In newsletter settings, if you choose a header of type Image (or another that does upload) and attach a file with the wrong file type (like PDF), you correctly see a message that says the file couldn’t be uploaded and needs to be of type png, jpg, or jpeg. But if you then attach a new file of type png/jpg/jpeg, the error message remains and says that the new file needs to be of the correct file type.

This fix clears out the error state when an attached file is removed or changed.

Fixes: CV2-3166.

## Type of change

## How has this been tested?

* Go to the newsletter settings page
* Choose a header type of Image and upload a PDF file
* Try to save and you'll get an error
* Click on the "X" icon in order to remove the uploaded file
* The error state in the upload component should be cleared

## Checklist

### DIFF
--- a/src/app/components/team/Newsletter/NewsletterComponent.js
+++ b/src/app/components/team/Newsletter/NewsletterComponent.js
@@ -100,6 +100,9 @@ const NewsletterComponent = ({
     } else if (!fileRequired) {
       setDisableSaveNoFile(false);
     }
+
+    // Reset any errors returned by the backend related to the previously uploaded file
+    setErrors({ ...errors, header_file: null, base: null });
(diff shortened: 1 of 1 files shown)


dariusk (approved):
Tested this locally and I noticed that if I upload a bad file and it renders an error, then I switch the header type to "None" and then back to "Image" header, the original file is still listed but without the error appended. I think that's okay but I wanted to flag it here (we would expect this behavior since the `useEffect` is triggered when `headerType` changes).

PR #1473, opened 2023-06-17
A couple of fixes for the newsletter RSS feed URL field.

caiosba:
## Description

1) Only prepend https:// when there is some value in the input field
2) Make sure that the "loading" state appears only when it should

Fixes CV2-3260.

## Type of change

## How has this been tested?

How to test this manually:

- Go to the newsletter settings page and enable RSS feed
- Add some invalid URL, like https://bla
- Save
- Reload the page
- Change the URL to some valid RSS feed URL
- Click "Load"
- It should load the articles successfully
- Save
- Reload the page
- Clear the RSS feed URL field
- Disable RSS feed
- Save
- Reload the page
- Enable RSS feed
- The RSS feed URL field should be empty

## Checklist


dariusk (approved):
Looks good, glad to have these checks in.

PR #1480, opened 2023-06-23
UI updates for custom lists

danielevalverde:
## Description

- Add an icon to the list page when the list is related to a feed
- Hovering over the icon tells which feeds this list is part of
- Add a new warning box to the confirmation dialog that is displayed when the custom list is saved
- Add unit tests

Reference: CV2-3196
## Type of change

## Checklist


dariusk (approved):
This looks really good! I pushed a commit to get rid of the eslint-disable directive about strings that lack descriptions. In one case, we were localizing a string that shouldn't be localized -- it's an enum, one of `'list'`, `'collection'`, or `'folder'`, and not user-facing. So I turned that back into an unlocalized string. The other one I added a description for.

The following is not a suggestion for changes, just something I noticed: the `eslint-disable relay/unused-fields` is still active in the file, and removing it shows a bunch of things that are unused explicitly in the component. But if we look closer we see that it's used in higher level objects we pass to children (for example, your `feeds` object is contained within `saved_search` which gets passed to other components). I don't think we need to fix this here, but it's more of a flag that these components overall could be architected a little more cleanly by doing things like passing a `feeds` object directly instead of a big object with a bunch of sub-objects inside it. But again, no need to fix in this PR.

brianfleming:
@danielevalverde is there a way to see this locally, in other words, get the data in a way that a list in my local workspace is part of a feed? [Maybe we update this?](https://meedan.atlassian.net/wiki/spaces/ENG/pages/1218248720/How+to+erase+your+local+database+and+create+fake+data)

brianfleming (approved):
Thanks for helping me get my local setup to see this @danielevalverde 
I added the style changes we talked about 🔆

PR #1481, opened 2023-06-27
Tags Component and Create Feed Integration

dariusk:
## Description

 * Create a CDS `Chip` component
 * Create a CDS `TagsList` component
   * For adding tags, this component uses `check-ui`'s `MultiSelector` and a `material-ui` `Popover`, just like selecting tags
   * There is a `readOnly` state for the tag list and it must be passed an array of strings (`tags`) and its associated `setTags` function
   * The `maxTags` property displays a `+N` chip where N is the remaining tags not shown over the `maxTags` limit. Hovering the chip shows a tooltip with the remaining tags listed (this is not used in the Create Feed UI but was in the spec for the `TagsList` component)
 * Create a "UI Sandbox" page only visible to superadmins at `/check/me/ui-sandbox`

In the sandbox:
 
image

In the create feed UI:

image

References: CV2-3191

## Type of change

## How has this been tested?

Self-tested -- unit tests coming soon!

I don't believe there is anything majorly tricky here. Worth looking at `Sandbox.js` just to see how the new UI sandbox works as a `QueryRenderer`.

## Checklist

- [X] I have removed the /* eslint-disable @calm/react-intl/missing-attribute */ from any files I've worked on and added a `description` prop to all `<FormattedMessage />` components that were missing it.


src/app/components/cds/buttons-checkboxes-chips/Chip.js
```diff
@@ -0,0 +1,44 @@
...
+};
+
+Chip.defaultProps = {
+  removable: false,
+  onRemove: () => {},
+};
+
+Chip.propTypes = {
+  onRemove: PropTypes.func,
+  removable: PropTypes.bool,
```
amoedoamorim, line 40: This seems a bit redundant. Can we get rid of the removable prop and consider removable if onRemove is passed?

dariusk: makes sense, I'll change that

amoedoamorim (approved):
All good!

src/app/components/cds/buttons-checkboxes-chips/Chip.js
dariusk: fixed in [d8dc2cf](https://github.com/meedan/check-web/pull/1481/commits/d8dc2cf87946b8dc67eca4da7ee0ce09e93fbe28)

dariusk:
Same unit tests failing here as on upstream, in `CreateFeed.js`, merging

PR #1468, opened 2023-06-13
[CV2-3243] Proposal to add classnames package to help manage css classes

brianfleming:
## Description

Discussion/Proposal PR to add the npm package [classnames](https://github.com/JedWatson/classnames) which is widely accepted as an easier way to manage multiple and conditional css classnames.
This PR:
- adds the package references
- updates the textfield component as an example on how to handle:
  - applying a global css class
  - handling multiple css module references

References: [CV2-3243](https://meedan.atlassian.net/browse/CV2-3243)

## Type of change

- [x] Performance improvement and/or refactoring (non-breaking change that keeps existing functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Automated test (add or update automated tests)

## How has this been tested?

manual testing

## Things to pay attention to during code review

- Take a look at the associated ticket to see the incorrect out of some class names from the textfield component:  [CV2-3243](https://meedan.atlassian.net/browse/CV2-3243)
- Link to the [classnames repo](https://github.com/JedWatson/classnames)
- Some other [interesting examples of its usage](https://mtm.dev/multiple-classnames-css-modules-react)

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code in hard-to-understand areas, if any
- [ ] I have made needed changes to the README
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If I implemented any new components, they are self-contained, their `propTypes` are declared and they use React Hooks and, if data-fetching is required, they use Relay Modern with fragment containers
- [ ] I have removed the /* eslint-disable @calm/react-intl/missing-attribute */ from any files I've worked on and added a `description` prop to all `<FormattedMessage />` components that were missing it.
- [ ] To the best of my knowledge, any new styles are applied according to the design system
- [ ] If I added a new external dependency, I included a rationale for doing so and an estimate of the change in bundle size (e.g., checked in https://bundlephobia.com/)

[CV2-3243]: https://meedan.atlassian.net/browse/CV2-3243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #1485, opened 2023-06-29
Add reset button to Select component

amoedoamorim:
## Description

This commit adds a reset button to the select component and related styles.
Reset button shows when Select is given an onRemove function prop.
Also sneaking some propTypes for SelectListQueryRenderer.

References: CV2-3311

## Type of change

## How has this been tested?

Tested manually. Select a value, then click the new X button to unselect.

Seem pretty straightforward. Not 100% sure about the best method to reset the native html select element, though. I guess this works but having multiple Selects would have ids clashing?

## Checklist


src/app/components/cds/inputs/Select.js
```diff
@@ -26,22 +31,38 @@ const Select = ({
...
-      </select>
-      <div className={inputStyles['input-icon-right-icon']}>
-        <ChevronDownIcon />
       </div>
+      { onRemove ?
+        <Tooltip title={<FormattedMessage id="select.removeSelection" defaultMessage="Undo selection" description="Tooltip for button on Select component to remove current selection" />}>
+          <IconButton
+            className={cx('select__clear-button', styles['select-clear-button'])}
+            onClick={() => {
+              document.getElementById('check-select__input').value = null;
```
amoedoamorim, line 58: what do people think of this? :thinking:

caiosba: Is there any race condition that can happen between this and the `onRemove` call?

src/app/components/cds/inputs/Select.js
```diff
@@ -26,22 +31,38 @@ const Select = ({
...
-      >
-        {children}
-      </select>
-      <div className={inputStyles['input-icon-right-icon']}>
-        <ChevronDownIcon />
       </div>
+      { onRemove ?
+        <Tooltip title={<FormattedMessage id="select.removeSelection" defaultMessage="Undo selection" description="Tooltip for button on Select component to remove current selection" />}>
+          <IconButton
+            className={cx('select__clear-button', styles['select-clear-button'])}
```
caiosba, line 56: What does `cx` do?

brianfleming: @caiosba this is a shortcut variable name for the new classNames package that we added to make it easier to manage multiple classnames, especially when they are applied with conditional logic. the cx is from the example repo...I think the author meant it as shorthand for classnames bind, since in this example alex is using both a global `classname select__clear-button` and a classname bound to a css modules file `select-clear-button`

you can check out some info here with links to the package:
[https://github.com/meedan/check-web/pull/1468](https://github.com/meedan/check-web/pull/1468)

caiosba: Ah cool, thanks! I'll start using it too.

src/app/components/cds/inputs/Select.js
caiosba: I think it's safer to keep it out and rely on the callback.

caiosba (commented):
Looks good! Pending decision on the `onClick` handler.

brianfleming (changes requested):
@amoedoamorim looking good. Two things. one of them is my oversight so apologies.

- [x] Can we change the "unselected" text to read `Select a custom filtered list...`
- [ ] When the user hits the new X button I think it should go back to the unselected state ^^ instead of blank

src/app/components/cds/inputs/Select.js
amoedoamorim: I don't really see a race condition, but I fixed this using a Ref. This old approach wouldn't work with multiple Selects in a page evidently.

src/app/components/cds/inputs/Select.js
```diff
@@ -1,56 +1,84 @@
...
-    <div className={`${inputStyles['input-container']} ${styles['select-container']}`}>
-      { iconLeft && (
-        <div className={inputStyles['input-icon-left-icon']}>
-          {iconLeft}
+}) => {
+  const inputRef = React.useRef(null);
+
+  return (
+    <div className={className}>
+      { console.log('inputProps', inputProps) /* eslint-disable-line */ }
```
caiosba: Ooops

caiosba (changes requested):

caiosba (approved):

brianfleming (approved):

PR #1488, opened 2023-06-29
Create feed error states

dariusk:
These are the errors on create. 
References: CV2-3192, CV2-3195

## Type of change

## How has this been tested?

Some unit testing, manual testing.

`TextField` has been changed to add a `suppressInitialError` prop. There's a new custom hook called `useEffectSkipFirst` that checks to see if a `ref` to the component itself has rendered and then acts as a passthrough to `useEffect`, so basically `useEffect` but does not trigger on the first render. There is now an `internalError` state that is either directly the same as `error` (because it initializes to `error` and then updates on subsequent renders) or is `false` on initial render and then equal to `error` on subsequent renders.

## Checklist


src/app/components/feed/SaveFeed.js
```diff
@@ -295,6 +307,27 @@ const SaveFeed = (props) => {
                   description="Title of the section where the publishing preferences such as licenses are selected"
                 />
               </div>
+              {
+                discoverableNoLicense && (
+                  <Alert
+                    id="save-feed__no-license-error"
+                    title={<span>Select a license in order to create and publish this shared feed.</span>}
```
amoedoamorim, line 314: Please don't forget to make it localizable

PR #1495, opened 2023-06-30
Updated WebStructured l10n resource

amoedoamorim:
## Description

The usual localization update. Just the resource file I just pushed to Transifex.
Since this is generated from code, I think we can discuss not commiting these files to the repo.  

Removed the remainder of PR template because it doesn't really apply here.

### DIFF
--- a/localization/transifex/WebStructured.json
+++ b/localization/transifex/WebStructured.json
@@ -180,16 +180,14 @@
 		"string": "Privacy policy"
 	},
 	"UserMenu.signOut": {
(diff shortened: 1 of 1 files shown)


caiosba (approved):

amoedoamorim:
Perhaps we can chat about not commiting this to the repo on a dev call in the near future :)

dariusk (approved):
I'm all ears about not committing this to the repo in the future!

brianfleming (approved):

PR #1504, opened 2023-07-06
replace more icons

brianfleming:
## Description

PR to continue to replace icons per [CV2-3355](https://meedan.atlassian.net/browse/CV2-3355)
Goal is to make only icon swap/addition and small style changes needed because of the difference in files, ie. keep the changes small.

References: [CV2-3355](https://meedan.atlassian.net/browse/CV2-3355)

## Type of change

## How has this been tested?

manual testing by visual verification

There should be very limited visual changes.

## Checklist

[CV2-3355]: https://meedan.atlassian.net/browse/CV2-3355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CV2-3355]: https://meedan.atlassian.net/browse/CV2-3355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/components/search/SearchField.js
```diff
@@ -8,23 +8,22 @@ import Grid from '@material-ui/core/Grid';
...
+import OpenInFullIcon from '../../icons/open_in_full.svg';
 
 const useStyles = makeStyles(theme => ({
   input: {
     borderRadius: 50,
     fontSize: 14,
   },
   inputInactive: {
-    border: `${borderWidthLarge} solid var(--grayBorderMain)`,
+    border: '2px solid var(--grayBorderMain)',
```
brianfleming, line 23: reviewers note. Eventually we will remove the vars in shared.js and update the search component with its own css file, so I made this change and at tiny step in that direction, but kinda unrealated to the icons slipped in while i was working

src/app/components/cds/inputs/TextField.test.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import VolumeUpIcon from '@material-ui/icons/VolumeUp';
-import VolumeOffIcon from '@material-ui/icons/VolumeOff';
 import { FormattedMessage } from 'react-intl';
 import TextField from './TextField';
+import VolumeUpIcon from '../../../icons/volume_up.svg';
```
brianfleming, line 5: @dariusk running into another testing issue after this change. I did some googling, but got lost in my less than average jest knowledge. Do you know of a change to make to help tests like this?

dariusk: Looking at these test failures now! I'll let you know.

src/app/components/cds/inputs/TextField.test.js
```diff
@@ -36,13 +35,13 @@ describe('<TextField />', () => {
 
   it('should render icons', () => {
     const textField = shallow(<TextField
-      iconLeft={<VolumeUpIcon />}
-      iconRight={<VolumeOffIcon />}
+      iconLeft={<VolumeUpIcon className="icon__vol-up"/>}
+      iconRight={<VolumeOffIcon className="icon__vol-off"/>}
```
dariusk, line 39: @brianfleming Basically the fix for the unit tests thing is to search for icons in test via class name rather than component name. This is because our custom icons are not actually rendered with a component name when we mock them, they are just raw SVG! So when we switch from a React component (such as a material-ui icon) to a bare SVG (like our custom icons), we can no longer use Jest's "find the component" shortcut selector.

brianfleming: ah, thats good to know, thanks @dariusk . will help me diagnose in the future

dariusk:
I'm seeing a message in the webpack build, `ERROR in [stylelint] Unexpected token '?'`

Is that happening for you or is it local to my machine? It doesn't seem to prevent compilation but it's new as of this branch.

brianfleming:
> I'm seeing a message in the webpack build, `ERROR in [stylelint] Unexpected token '?'`
> 
> Is that happening for you or is it local to my machine? It doesn't seem to prevent compilation but it's new as of this branch.

yeah, you are right, I am seeing it locally, but doesnt stop the compile. its rather cryptic message

dariusk:
I'll try and squash it

dariusk:
Fixed! For whatever reason `stylelint` suddenly started needing the `.stylelintrc` file to be strict JSON rather than a JS-object-like-thing. I ran it through `jsonlint` and found the couple places where there was a stray comma that there shouldn't be.

dariusk (approved):

caiosba:
> Fixed! For whatever reason `stylelint` suddenly started needing the `.stylelintrc` file to be strict JSON rather than a JS-object-like-thing. I ran it through `jsonlint` and found the couple places where there was a stray comma that there shouldn't be.

I wonder if it's related to the recent upgrade: https://github.com/meedan/check-web/commit/7189b757799490487088836dadb20d0339e52bcc

brianfleming:
> Fixed! For whatever reason `stylelint` suddenly started needing the `.stylelintrc` file to be strict JSON rather than a JS-object-like-thing. I ran it through `jsonlint` and found the couple places where there was a stray comma that there shouldn't be.

wow, nice catch. Suddenly being held to a higher standard is a great reason for a new error! 😅

dariusk:
@caiosba That's definitely the reason!

PR #1520, opened 2023-07-14
Improve LimitedTextArea and TextField perf

dariusk:
`TextField` and `TextArea` (and `LimitedTextArea`, which calls them) were rerendering their parent components on every keystroke. We normally solve this by switching from `onChange` to `onBlur` for setting values, but a problem arises with `LimitedTextArea`, which requires its own state to change on every keystroke (in order to display the number of remaining characters). If we did the normal `onBlur` solution then the character count would only update on blur.

`TextField` and `TextArea` now use the `forwardRef` pattern. This lets us set a `ref` on them just like we might do on an `input` element (by default in React, `ref` only works on basic HTML elements and not components -- using `forwardRef` allows us to use `ref` on components as well). Other than wrapping these in `forwardRef` and adding the forwarded `ref` to the `textarea` and `input` HTML elements, no changes were made to the functionality of these components.

So they behave exactly how they did before, they just also have the ability to take a `ref`. Importantly this means we have at least 3 options when using `TextField` and `TextArea`:

 - use a standard onChange pattern with `useState` where you pass `value={myText}` and `onChange={e => setMyText(e.target.value)}` (rerenders the parent every keystroke)
 - use an onBlur pattern with `useState` where you `value={myText}` and `onBlur={e => setMyText(e.target.value)}` (rerenders the parent on blur)
 - and, new as of this commit, declare `const inputRef = React.useRef()` in the parent component, pass `ref={inputRef}` on `TextField` or `TextArea`, and then refer to `inputRef.current.value` in the parent to get access to the value in the text component (NEVER rerenders the parent and in fact just behaves like an uncontrolled component). If you need a local state tied to the value of the input (like displaying character count) you can avoid a parent rerender by creating a local `useState` copy of the `ref` value and updating it on keystrokes but only ever referring to the state in the scope of the current component (so the current component rerenders but the parent doesn't, much more efficient). You can see how it's done in `LimitedTextArea.js` for an example of implementing this style of input.

Meanwhile, in addition to implementing the third solution above, `LimitedTextArea` gets a change in its `useEffect` for error handling. We do want to check for errors on every rerender of `LimitedTextArea` so the `useEffect` is correct here, but previously we were constantly updating the parent element's `onErrorTooLong` on every keystroke! So even with the `ref` stuff implemented we were *still* rerendering the parent (we had been rerendering its parents *twice* on every keystroke actually, the first case taken care of by using `ref` but the second case remained). This commit makes a change where we only update the error state on the parent when we *transition* from one error state to another. So ultimately, what triggers a rerender of the parent of `LimitedTextArea` is: an onBlur event, or a change in error state.

Other changes:

 - add `LimitedTextArea` to the UI sandbox
 - make `NewsletterStatic` update article values onBlur instead of onChange
 - updated the keys in `NewsletterStatic` to use index numbers and squashed the linter error - normally this is unadvised but in this case it's an array we only append items to or remove items from the end so it's fine, and there was a little hiccup where the keys were all `undefined` until the data loaded from the API so this squashes that warning
 - fix some unit test warnings

References: CV2-3373

## Type of change

- [X] Automated test (add or update automated tests)

## How has this been tested?

Unit tests on the core components continue to pass, and I did extensive manual testing on the Newsletter settings page to make sure everything worked as needed. There's not actually any new or changed functionality on the components so tests continuing to pass for the text components should suffice.

There's not any particular code I want to call out - my hope is that the description above makes sense and we can use this as a reference for future problems where need to streamline inputs.

The easiest way to see this in action is to go to the Newsletter settings page in production and see how slow it is, then check it on your local machine running this branch and see that it's very fast.

## Checklist

### DIFF
--- a/src/app/components/Sandbox.js
+++ b/src/app/components/Sandbox.js
@@ -49,8 +49,6 @@ const SandboxComponent = ({ admin }) => {
   const [switchesHelp, setSwitchesHelp] = React.useState(Boolean(false));
   const [switched, setSwitchExample] = React.useState(Boolean(false));
   const [limitedText, setLimitedText] = React.useState('Hello this is the initial limited text state');
-  // eslint-disable-next-line
-  console.log('~~~RENDER');
 
   const [switchLabelPlacement, setSwitchLabelPlacement] = React.useState('top');
   const onChangeSwitchLabelPlacement = (event) => {
--- a/src/app/components/Sandbox.js
+++ b/src/app/components/Sandbox.js
@@ -17,6 +17,7 @@ import AddIcon from '../icons/settings.svg';
 import ListIcon from '../icons/list.svg';
 import FigmaColorLogo from '../icons/figma_color.svg';
 import Card from './cds/media-cards/Card.js';
+import LimitedTextArea from './layout/inputs/LimitedTextArea';
 
 const SandboxComponent = ({ admin }) => {
   const isAdmin = admin?.is_admin;
@@ -47,6 +48,9 @@ const SandboxComponent = ({ admin }) => {
   const [switchesDisabled, setSwitchesDisabled] = React.useState(Boolean(false));
   const [switchesHelp, setSwitchesHelp] = React.useState(Boolean(false));
   const [switched, setSwitchExample] = React.useState(Boolean(false));
+  const [limitedText, setLimitedText] = React.useState('Hello this is the initial limited text state');
+  // eslint-disable-next-line
+  console.log('~~~RENDER');
 
   const [switchLabelPlacement, setSwitchLabelPlacement] = React.useState('top');
   const onChangeSwitchLabelPlacement = (event) => {
@@ -72,6 +76,7 @@ const SandboxComponent = ({ admin }) => {
   const onChangeButtonTheme = (event) => {
     setButtonTheme(event.target.value);
   };
+
   return (
     <div className={styles.sandbox}>
       <h5>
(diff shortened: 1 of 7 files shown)


src/app/components/cds/inputs/SwitchComponent.js
```diff
@@ -69,7 +69,7 @@ SwitchComponent.propTypes = {
   disabled: PropTypes.bool,
   label: PropTypes.object,
   labelPlacement: PropTypes.oneOf(['bottom', 'end', 'start', 'top']),
-  helperContent: PropTypes.string,
+  helperContent: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
caiosba, line 72: Nice approach!

src/app/components/Sandbox.js
```diff
@@ -47,6 +48,9 @@ const SandboxComponent = ({ admin }) => {
   const [switchesDisabled, setSwitchesDisabled] = React.useState(Boolean(false));
   const [switchesHelp, setSwitchesHelp] = React.useState(Boolean(false));
   const [switched, setSwitchExample] = React.useState(Boolean(false));
+  const [limitedText, setLimitedText] = React.useState('Hello this is the initial limited text state');
+  // eslint-disable-next-line
+  console.log('~~~RENDER');
```
caiosba, line 53: Question: Is this intended because this is a sandbox?

caiosba (approved):
I just have one question but good to go! Thanks for the great documentation in the description.

src/app/components/Sandbox.js
dariusk: oops, will remove

PR #1534, opened 2023-07-24
Icons and improvements to Main Button component

brianfleming:
## Description
Just continuation on other work to get the app to be more like the design system. Tickets referenced below

- Continuing the process of replacing icons
- Updating the Main Button component to remove material design since we don't need it for this application
- Updating some translation strings to add descriptions
- Replace some IconButton components with Main Button center icon variant to match design system
- Replace some instances of `<Typography` with plain html

References: [CV2-3355](https://meedan.atlassian.net/browse/CV2-3355) [CV2-3353](https://meedan.atlassian.net/browse/CV2-3353)

## Type of change

## How has this been tested?

manual visual comparisons

should be pretty straightforward

## Checklist

[CV2-3355]: https://meedan.atlassian.net/browse/CV2-3355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CV2-3353]: https://meedan.atlassian.net/browse/CV2-3353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/components/media/MediaActionsMenuButton.js
```diff
@@ -127,20 +127,23 @@ class MediaActionsMenuButton extends React.PureComponent {
...
     ));
 
     return menuItems.length ? (
-      <div>
-        <IconButton
+      <>
+        &nbsp;
+        <ButtonMain
           id="media-actions-menu-button__icon-button"
-          tooltip={<FormattedMessage id="mediaActions.tooltip" defaultMessage="Item actions" />}
```
brianfleming, line 139: @amoedoamorim I wanted to call your attention to this for some guidance. Before this change, this tooltip attribute was not doing anything, if I looked at the rendered code, it was outputting an object string, so even if it was functioning as a tooltip it wouldn't display correctly:

Screenshot 2023-07-24 at 2 31 57 PM

Regardless, as you can see i changed it to a title attribute on the Main Button component.

### The question:
Using this format message version it removed the strings from the json file. I just wanted to check if thats ok, or if there is something else I should do? I will note the json files in another comment so you can find them

localization/react-intl/src/app/components/media/MediaActionsMenuButton.json
```diff
@@ -1,30 +1,32 @@
...
+    "description": "Menu item to move the current item to the spam list",
     "defaultMessage": "Mark as Spam"
   },
   {
     "id": "mediaActions.history",
+    "description": "Menu item to view the history of changes to the item",
     "defaultMessage": "Item history"
-  },
-  {
-    "id": "mediaActions.tooltip",
```
brianfleming, line 27: @amoedoamorim here is one that was removed as noted in my other comment

amoedoamorim: yeah, that's probably happening because of the inline passing of the message descriptor object. Please use the `defineMessages` import from `react-intl` to generate the strings correctly for translation!

brianfleming: got it! thanks @amoedoamorim , I updated the strings

brianfleming:
Anyone have insight on those test failures? Doesn't seem like my changes to should result in those?

brianfleming:
> Anyone have insight on those test failures? Doesn't seem like my changes to should result in those?

oh disregard. I guess the branch just needed to be caught up

brianfleming:
Since this is just an ongoing project, I think this PR is big enough to be at a merge/stopping point if I could get a review @amoedoamorim and/or @dariusk 
muito apreciado

src/app/components/LanguageSwitcher.js
```diff
@@ -93,9 +93,7 @@ const LanguageSwitcher = (props) => {
                 </div>
                 { props.onSetDefault && languageCode !== primaryLanguage ?
                   <div>
-                    <IconButton onClick={handleClick}>
-                      <MoreVertIcon />
-                    </IconButton>
+                    <ButtonMain iconCenter={<MoreVertIcon />} variant="outlined" size="default" theme="text" onClick={handleClick} />
```
amoedoamorim, line 96: Just out of curiosity, any reasoning behind explicitly passing size=default, since it's the default value for that prop?

brianfleming: well only for 2 reasons really. 
1. for clarity sake when other devs look at the code there is less "hidden"
2. I was looking at other places where we might want to use the `ButtonMain` component that are standalone, meaning we might want to be able to set no size, no theme, etc and then style for that one location. Using css modules makes its harder if those styles are applied (ie. overriding parent styles from a child component). So there might be a case where there is no default prop set when empty, its just empty. 

That being said, I agree at the current state of this PR its largely unnecessary, but harmless in my opinion, just superfluous.

amoedoamorim (approved):
:LGTM:

PR #1543, opened 2023-07-27
Fixing "refresh URL" integration test

caiosba:
## Description

This commit fixes the "refresh URL" integration test. The URL it was using is not valid anymore, so I replaced by another one.

## Type of change

## How has this been tested?

All integration tests should pass.

I would love to find another testing URL that provided the same functionality (a different page title every time we refresh it).

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ before_script:
 jobs:
   include:
     - stage: tests
-      # if: branch = develop OR branch = master
+      if: branch = develop OR branch = master
       name: integration-and-unit-tests
       script:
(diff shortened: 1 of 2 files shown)


dariusk:
Man, I just don't love this sort of thing because ultimately if `time.is` goes down our integration tests fail. Could we ping something like `[local api host]:3000` and just grab the `Welcome to Check. This is the root of the core API service.` page instead? At least if _that_ service is down, it's a problem we need to fix.

amoedoamorim (approved):

PR #1549, opened 2023-08-02
New core list: Unmatched media

caiosba:
## Description

* New core list "Unmatched media", accessible from the left sidebar, right below "Suggestions", with a custom icon and only displayed if similarity is enabled
* Other filters can be applied, but they can just be saved as a new custom list
* When a suggestion is reject, there is no modal to choose a destination folder anymore: just a success message is displayed
* When suggestions are bulk-rejected, there is no modal to choose a destination folder anymore: just a confirmation modal
* When detaching an item, there is no modal to choose a destination folder anymore: just a success message is displayed

Reference: CV2-3434.

## Type of change

## How has this been tested?

There were integration tests for the rejection/detach workflow, I'm fixing them.

On the "Unmatched media" page, I didn't care about the logic used to display the "apply" and "save" buttons and didn't care about displaying the "unmatched" filter, since those things are going to be handled for all core lists in separate tickets that were already created.

## Checklist


dariusk (commented):
Looks good overall! I tested the feature and it seems to work as intended.

Couple code questions:

Is there a reason we don't count this as an `isSpecialPage` on `src/app/components/search/SearchFields/index.js` [line 289](https://github.com/meedan/check-web/blob/722583088e1db076423e2c5b43f86ec21d14c08d/src/app/components/search/SearchFields/index.js#L289)? I'm not really sure what that does, I just noticed it as a pattern where we might want to include it.

caiosba:
> Looks good overall! I tested the feature and it seems to work as intended.
> 
> Couple code questions:
> 
> Is there a reason we don't count this as an `isSpecialPage` on `src/app/components/search/SearchFields/index.js` [line 289](https://github.com/meedan/check-web/blob/722583088e1db076423e2c5b43f86ec21d14c08d/src/app/components/search/SearchFields/index.js#L289)? I'm not really sure what that does, I just noticed it as a pattern where we might want to include it.

Good question, Darius! I didn't add it there because that logic is used to just hide the "number of suggestions" filter.

dariusk (approved):

amoedoamorim (approved):

PR #1555, opened 2023-08-05
Fixing integration test.

caiosba:
## Description

There was one integration test failing (`should pin and remove similarity items`). It started to fail after a `gap: 4px` was added to a `div` (https://github.com/meedan/check-web/blob/da497f2373eb2163d2e859f6d4dd2361a126526b/src/app/components/media/MediaCardLargeActions.js#L141). It just happens in the Chrome version used by Chromedriver... doesn't happen in the latest Chrome version. Also, it just happens when that `div` is empty.

The fix here is to simply not even show that `div` when the media type is "claim", since in that case the `div` is empty.

## Type of change

## How has this been tested?

Integration tests should pass.

Welcome other ideas to fix this.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ before_script:
 jobs:
   include:
     - stage: tests
-      # if: branch = develop OR branch = master
+      if: branch = develop OR branch = master
       name: integration-and-unit-tests
       script:
       - docker-compose exec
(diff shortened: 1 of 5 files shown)


brianfleming:
Interesting. I believe I added that gap during a change to the new button to space them out a bit without having to touch too many other things.

@caiosba curious if I am looking at the right place. Is the chromediver version this? from the test output in travis?
`(Session info: chrome=84.0.4147.105)`

If so, its even more curious, because according to caniuse. support for gap was added right around that version:
Screenshot 2023-08-07 at 9 25 16 AM

Regardless, good detective work and fix @caiosba

caiosba:
Thanks @brianfleming , yes, that is the Chrome version used in the integration tests. And actually `gap` works... it's just this buggy behavior in that specific case: when there are no children, the component gets the full height.

brianfleming (approved):
What a weird one! 🙌

PR #1559, opened 2023-08-07
[Hotfix] - Reenable paginated media

dariusk:
## Description

Same as #1558 but this PR is to `master` for the hotfix so we don't push other changes from QA to live.


amoedoamorim (approved):

dariusk:
Merging to develop instead

PR #1568, opened 2023-08-10
CV2-3165: Add Published default list

amoedoamorim:
## Description

This PR adds a new default list: Published (as in published reports)

## Type of change

## How has this been tested?

Tested manually that there's a new list heading Published on the sidebar and it opens the Published list page, with the "Report is published" filter active.
As far as I understand no new integration tests are needed, there are enough of them testing the search/filters mechanism.

- Etc.

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19837,7 +19837,7 @@
     "requireindex": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/requireindex/-/requireindex-1.1.0.tgz",
-      "integrity": "sha1-5UBLgVV+91225JxacgBIk/4D4WI="
+      "integrity": "sha512-LBnkqsDE7BZKvqylbmn7lTIVdpx4K/QCduRATpO5R+wtPmky/a8pN1bO2D6wXppn1497AJF9mNjqAXr6bdl9jg=="
     },
     "resize-observer-polyfill": {
       "version": "1.5.1",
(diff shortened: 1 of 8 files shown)


package-lock.json
caiosba, line 1: Question: Please review if the changes here in `package-lock.json` are really needed.

src/app/components/drawer/Projects/ProjectsComponent.js
```diff
@@ -298,6 +299,20 @@ const ProjectsComponent = ({
           </ListItem>
         }
 
+        { team.smooch_bot &&
```
caiosba, line 302: Request: Remove this condition please. We don't need to have the condition here because workspaces can have published reports by just using the API (as the ticket specified: "Always display this list").

src/app/components/search/SaveList.js
caiosba, line 1: Question: Please confirm: You don't need any logic in the `handleSave` method( like this, for example: https://github.com/meedan/check-web/blob/develop/src/app/components/search/SaveList.js) because the filter is exposed, right?

caiosba (changes requested):
I tried it locally and works as expected! A couple of questions and really only one requested change.

package-lock.json
amoedoamorim: Not needed. Should be no actual change, actually. The package version didn't change. Just the integrity hash from SHA-1 to SHA-512. Should I remove this?

src/app/components/drawer/Projects/ProjectsComponent.js
amoedoamorim: ok

package-lock.json
caiosba: SHA-512 means the newer NPM version, right? So I think we want SHA-512 wherever possible.

src/app/components/search/SaveList.js
amoedoamorim: Yeah, I tried saving and it was not needed. I think this is because it already comes from the query object initialized in the list component `Published.js`. So I guess that was only needed for projects actually, because the project filter was implicit when loading a project and that had to be made explicit when it became a saved search. I'm not sure about the other cases that are handled there.

caiosba (approved):
Thanks for addressing the review comments!

PR #1579, opened 2023-08-17
Avoid displaying second-page tag as "property deleted" in filter

amoedoamorim:
## Description

Due to tags pagination, a tag used in a query can be on a not loaded yet page.

Merge `query.tags` with plainTagsText even if they're not loaded from graphql.

References CV2-3336

## Type of change

## How has this been tested?

Tested manually that applying a tag filter selecting tags that are loaded after scrolling (on load more). After the filter is applied I verified that the filter is displayed correctly, instead of showing "Property deleted" for the not loaded tags.

- Etc.

## Checklist


caiosba (approved):
Pair-reviewed this with @amoedoamorim  on a call right now - clarified that using `new Set` will already filter potential duplicated. Good catch @amoedoamorim !

PR #1600, opened 2023-08-29
add min and max height on textArea Component

danielevalverde:
## Description

- fix: do not apply 'overflow-y: hidden' and 'resize: none' properties on textArea if autogrow is false
- and add prototypes

Reference: CV2-3180
## Type of change

## Checklist


src/app/components/cds/inputs/TextField.js
```diff
@@ -81,6 +82,7 @@ const TextField = React.forwardRef(({
                 [styles.outlined]: variant === 'outlined',
                 [styles['input-icon-left']]: iconLeft,
                 [styles['input-icon-right']]: iconRight,
+                [styles['textarea-autoGrow']]: autoGrow,
```
brianfleming, line 85: @danielevalverde since you are adding this class to the `<textarea` part of this component, the css definition below, needs to target that as well.

src/app/components/cds/inputs/TextField.module.css
```diff
@@ -70,6 +70,9 @@
     textarea {
       max-width: 100%;
       min-height: 3.5em;
+    }
+
+    .textarea-autoGrow {
```
brianfleming, line 75: this css definition needs to be `&.textarea-autoGrow` inside of the `textarea {` portion above...so inside of lines 71-72, so it gets applied to the right spot in the dom

src/app/components/cds/inputs/TextArea.js
```diff
@@ -1,20 +1,36 @@
...
   };
 
-  return <TextField textArea ref={ref} {...inputProps} style={{ height }} onChange={handleChange} />;
+  return <TextField textArea autoGrow ref={ref} {...inputProps} style={{ height }} onChange={handleChange} />;
 });
 
+TextArea.defaultProps = {
+  autoGrow: true,
+  maxHeight: null,
+  minHeight: 65,
```
brianfleming, line 27: I'm not sure this is the best way to approach this. The reason is, the TextField css field css file has a min-height set for text area at `min-height: 3.5em;` and I think its a bad idea to have two different ways of setting this. Some suggestions are:
- in the js look at the css value, I think its `getComputedStyle`
- go down to 1 set in the js, and make it required so its clear to developers whats going on
- other? any comment here @amoedoamorim ?

brianfleming (changes requested):
@danielevalverde I put some inline comments in, but also:

- [ ] when I use the sandbox with all the defaults, the text area still initially "shrinks" when I type, it should only grow past its min-height
- [ ] when I set autoGrow to false, the css class value is not actually passed to the TextField component, so even when you correct my inline comments about correctly applying the css definition in the right place, the css class never gets actually applied

So I suggest verifying things on the sandbox as I have done, its pretty useful!
let me know if I can help or add more clarity to my comments.

src/app/components/cds/inputs/TextArea.js
amoedoamorim: This might be coming a bit late, but I wonder if this could be solved with an almost css-only approach in which autoGrow would mean applying either styles (with or without maxHeight). Wouldn't it be cleaner? Am I missing something here?

src/app/components/cds/inputs/TextArea.js
```diff
@@ -1,20 +1,37 @@
...
-    if (autoGrow) {
-      setHeight(`${event.target.scrollHeight}px`);
+    const minHeight = minHeightProp || event.target.offsetHeight;
+    const newHeight = `${event.target.scrollHeight}px`;
+    if (autoGrow && event.target.scrollHeight >= minHeight && (!maxHeight || event.target.scrollHeight <= maxHeight)) {
+      setHeight(newHeight);
     }
   };
 
-  return <TextField textArea ref={ref} {...inputProps} style={{ height }} onChange={handleChange} />;
```
caiosba, line 17: What happens if the user sends a custom `style`? I think we need to merge it here:

```
const customStyle = inputProps.style || {};
style={{ ...customStyle, height }}
```

src/app/components/cds/inputs/TextArea.js
```diff
@@ -1,20 +1,37 @@
...
 
+
   const handleChange = (event) => {
-    if (autoGrow) {
-      setHeight(`${event.target.scrollHeight}px`);
+    const minHeight = minHeightProp || event.target.offsetHeight;
+    const newHeight = `${event.target.scrollHeight}px`;
+    if (autoGrow && event.target.scrollHeight >= minHeight && (!maxHeight || event.target.scrollHeight <= maxHeight)) {
+      setHeight(newHeight);
     }
```
caiosba, line 20: What happens if the user defines a custom `onChange`? I think we need to call it here:

```
const handleChange = (event) => {
  // [...] Current code here
  if (inputProps.onChange) {
    inputProps.onChange();
  }
}
```

caiosba (changes requested):

src/app/components/cds/inputs/TextArea.js
```diff
@@ -1,12 +1,40 @@
...
-  <TextField textArea ref={ref} {...inputProps} />
-));
+}, ref) => {
+  const [height, setHeight] = useState('auto');
+
+  const handleChange = (event) => {
+    const minHeight = minHeightProp || event.target.offsetHeight;
+    const newHeight = `${event.target.scrollHeight}px`;
+    if (inputProps.onChange) {
+      inputProps.onChange();
```
caiosba, line 18: Not sure if `event` must be passed to `onChange`... I think so, please test.

brianfleming (approved):
Thanks for all the hard work @danielevalverde its looking good!

mario

PR #1617, opened 2023-09-12
CV2-3435 [full ci] adds support for presto

DGaffney:
## Description

Adds support for presto into check-web

References: 3435

## Type of change

## How has this been tested?

Currently not tested - just adding modifications as per conversation with @caiosba 

My first push into check-web so please let me know!

## Checklist


docker-test.yml
```diff
@@ -38,6 +38,10 @@ services:
     environment:
       BOILERPLATE_ENV: test
       ALEGRE_PORT: 3100
+  presto:
+    environment:
+      BOILERPLATE_ENV: test
+      ALEGRE_PORT: 8000
```
caiosba, line 44: `ALEGRE_PORT`? 8000 is the Presto port, no?

DGaffney: whoops! Fixed

DGaffney:
@dmou & @caiosba this is now failing because, I think, there's no `develop` branch in `presto` - but that is also (as I understand) a feature and not a bug of the new setup for this codebase (i.e. we are only building off master now). What should we do here? What can I do to resolve what seems to be a tension with the rest of the setup? Should we just `echo 'master'` instead of `echo 'develop'` in the `.travis.yml`?

dmou:
> > @dmou & @caiosba this is now failing because, I think, there's no `develop` branch in `presto` - but that is also (as I understand) a feature and not a bug of the new setup for this codebase (i.e. we are only building off master now). What should we do here? What can I do to resolve what seems to be a tension with the rest of the setup? Should we just `echo 'master'` instead of `echo 'develop'` in the `.travis.yml`?
> 
> Eventually we're going to have a `develop` branch, no?

uh we had determined like a year ago that we were moving toward single branch deployments, which would be a master/main branch, and not a develop/qa branch, so that we can have actual trunk based development. are we walking that back?

caiosba:
> > > @dmou & @caiosba this is now failing because, I think, there's no `develop` branch in `presto` - but that is also (as I understand) a feature and not a bug of the new setup for this codebase (i.e. we are only building off master now). What should we do here? What can I do to resolve what seems to be a tension with the rest of the setup? Should we just `echo 'master'` instead of `echo 'develop'` in the `.travis.yml`?
> > 
> > 
> > Eventually we're going to have a `develop` branch, no?
> 
> uh we had determined like a year ago that we were moving toward single branch deployments, which would be a master/main branch, and not a develop/qa branch, so that we can have actual trunk based development. are we walking that back?

Oh, yeah... I forgot about it. @DGaffney , since Check Web (and all other softwares of the suite) follow the "old" deployment way, I suggest we just fallback to `master` or `main`when there is no `develop`.

DGaffney:
Sounds good @dmou and @caiosba - thank you for resolving!

caiosba (approved):
Good to go! Let's please merge this one last, @DGaffney (after Check API and Alegre).

PR #1624, opened 2023-09-13
alternative approach to an autogrow textarea CV2-3180

brianfleming:
## Description

This is an alternative approach to an autogrow text area.

At its heart, what this does is add a "ghost" element that overlays the text area and copies its content, the wrapper is then set to be the height of the content. The "ghost" content pushes the text area to grow to match its size. 

It uses the "rows" attribute as a way to set the minimum height.

### Missing
Just need to add in the max height concept, but thats pretty trivial

References: CV2-3180

## Type of change

## How has this been tested?

- Etc.

## Checklist

### DIFF
--- a/src/app/components/cds/inputs/TextArea.js
+++ b/src/app/components/cds/inputs/TextArea.js
@@ -1,4 +1,3 @@
-/* eslint-disable */
 // DESIGNS: https://www.figma.com/file/rnSPSHDgFncxjXsZQuEVKd/Design-System?type=design&node-id=3606-26274&mode=design&t=ZVq51pKdIKdWZicO-4
 import React from 'react';
 import PropTypes from 'prop-types';
@@ -10,7 +9,9 @@ const TextArea = React.forwardRef(({
   maxHeight,
   ...inputProps
 }, ref) => {
-  const handleChange = (event) => event.target.parentNode.dataset.replicatedValue = event.target.value;
+  const handleChange = (event) => {
+    event.target.parentNode.setAttribute('data-replicated-value', event.target.value);
+  };
 
   const customStyle = inputProps.style || {};
   return <TextField textArea autoGrow={autoGrow} ref={ref} {...inputProps} style={{ ...customStyle }} rows={rows} onInput={handleChange} />;
--- a/src/app/components/cds/inputs/TextArea.js
+++ b/src/app/components/cds/inputs/TextArea.js
@@ -6,7 +6,6 @@ import TextField from './TextField';
 const TextArea = React.forwardRef(({
   autoGrow,
   rows,
-  maxHeight,
   ...inputProps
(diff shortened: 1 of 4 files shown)


brianfleming:
need some help with removing the eslint disable and to make it "real" code

src/app/components/cds/inputs/TextArea.js
```diff
@@ -1,43 +1,31 @@
...
-    if (event.target.value === '') {
-      setHeight('auto');
-    } else if ((autoGrow && event.target.scrollHeight >= minHeight) && (!maxHeight || event.target.scrollHeight <= maxHeight)) {
-      setHeight(newHeight);
-    }
-    if (inputProps.onChange) {
-      inputProps.onChange(event);
-    }
-  };
+  const handleChange = (event) => event.target.parentNode.dataset.replicatedValue = event.target.value;
```
brianfleming, line 13: This is obviously the code the requires the eslint disable, a smart meedan engineer should be able to resolve it better than I.

src/app/components/cds/inputs/TextArea.js
```diff
@@ -1,43 +1,31 @@
...
-  const handleChange = (event) => {
-    const minHeight = minHeightProp || 0;
-    const newHeight = `${event.target.scrollHeight}px`;
-    // if the input is empty reset the height
-    if (event.target.value === '') {
-      setHeight('auto');
-    } else if ((autoGrow && event.target.scrollHeight >= minHeight) && (!maxHeight || event.target.scrollHeight <= maxHeight)) {
-      setHeight(newHeight);
-    }
-    if (inputProps.onChange) {
```
brianfleming, line 23: not sure if somethign like this was still needed since it was changed from an `onChange` to an `onInput`

dariusk:
Your linter problems caused me to go down a javascript/DOM nerd hole.

There is a really silly interaction between the `dataset` DOM API and a linter rule that is 99% of the time the correct thing to do. Basically: when you have a function that is passed `event` and you set the data attribute using `event.target.node.dataset = blah`, that is correct and works! But the linter rule doesn't want you writing over any arguments passed in a function. The solution here is to use the alternative way of setting the attribute: calling `setAttribute` on the node.

This is really just linter pedantry clashing with a convenient but non-idiomatic DOM API feature and it's solved by not using `dataset`

dariusk:
(There was another error about function arguments returning things but that one was pretty obvious to solve; the linter error I'm addressing is the error that you get *after* you fix the obvious problem)

src/app/components/cds/inputs/TextArea.js
dariusk: yup I'll make a commit

dariusk (approved):
Seems to work pretty well and now lint-free!

PR #1625, opened 2023-09-13
move setting maxheight on the textarea component CV2-3180

brianfleming:
## Description

Setting of the maxheight option on the text area component resulted in the ghost element to continue to grow. This PR moves the logic to the wrapper div in order for everything to change height at the same time.

References: CV2-3180

## Type of change

## How has this been tested?

manual sandbox testing

### BEFORE
Screenshot 2023-09-13 at 2 55 04 PM

### AFTER
Screenshot 2023-09-13 at 3 24 35 PM

## Checklist

### DIFF
--- a/src/app/components/Sandbox.js
+++ b/src/app/components/Sandbox.js
@@ -374,6 +374,7 @@ const SandboxComponent = ({ admin }) => {
                   <option value="none">none</option>
                   <option value="48px">48px</option>
                   <option value="96px">96px</option>
+                  <option value="180px">180px</option>
                 </Select>
               </li>
               <li>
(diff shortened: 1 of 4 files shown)


amoedoamorim (approved):
Tried locally and seems to work well, though I'm still not fully aware of the engineering of the solution here

brianfleming:
> Tried locally and seems to work well, though I'm still not fully aware of the engineering of the solution here

I'm sure you wouldn't accept the answer of "magic"...but, really it works because we now have 2 elements that need to obey the max-height of the component, whereas the previous implementation only took into account the textarea. So the other element would keep growing, thereby pushing down the overall container height.

PR #1626, opened 2023-09-13
Feature/3569 pagination tag settings

dariusk:
This was implemented following the general patterns outlined in https://meedan.atlassian.net/wiki/spaces/ENG/pages/1185316865/How+to+paginate+in+Relay+1.7

The following things were changed from the spec in Jira, or were design decisions I made while implementing since they weren't specified:

 * search for a tag name now happens on blur or on enter key press, since we don't want to make an API call on every keypress
 * no more sorting of tags. since we aren't preloading all items by definition, clientside sort no longer makes sense and couldn't work. Backend sort could be implemented, perhaps alongside `keyword` on `TagTextConnection`. Maybe we should make a new ticket for this.
 * I did not extract the paginator as a standalone component as this needs more design work to be generally useful.

I ended up using `relay.refetchConnection()` to re-query the database after we do add/delete actions that mess up the pagination (such as adding/removing items in arbitrary places, as opposed to just pushing/popping from an array). More info here on the internal wiki: https://meedan.atlassian.net/wiki/spaces/ENG/pages/1185316865/How+to+paginate+in+Relay+1.7#Updating-lists-of-items

References: CV2-3569

## Type of change

- [X] Automated test (add or update automated tests)

## How has this been tested?

Integration tests for tag-based functionality continue to pass, and extensive local testing.

At some point I would like to abstract this into its own `Paginator` component: https://github.com/meedan/check-web/compare/feature/3569-pagination-tag-settings?expand=1#diff-96770a52089c3b97afdd0e1516b444a4181d8d4ac2eaf78443fad80e52401122R102-R157

@brianfleming - the paginator is not exactly the one we use on the All Items page. It's actually using more of our new CDS components, and looks pretty similar though not identical. It would be good if we could nail down a standalone component for pagination and then just design based off that.

image

## Checklist


amoedoamorim (approved):
Looks great. Extracting the pagination navigation to a component is a great idea.

PR #1627, opened 2023-09-13
Hotfix: initial height of pre-populated TextArea

dariusk:
We need to set the height of `TextArea` to the correct value if it has text already in it. What I do here is repeat what we normally do to get the correct height but make sure to do it on initial render using the `forwardRef` to access the element instead of via `event`.

I saved a long newsletter item and reloaded the page and this is what the initial render looks like (correct height):

image

I also added a long text to the limited textarea in the sandbox that is easy to see as well.

## Type of change

Notice that the new code is copying [what we already do](https://github.com/meedan/check-web/blob/47a846cca1cfbe4bba13cff935c497465b729f1c/src/app/components/cds/inputs/TextArea.js#L18-L23), just using the `ref` to access the DOM element instead of `event` (since this is not triggered on an event, just initial render).

## Checklist


caiosba (approved):
Tested locally and works as expected! Code looks good too. Thanks Darius!

PR #1643, opened 2023-09-27
Sentry screening 2023-09-27

dariusk:
This commit fixes a few small errors we discovered during our error screening process.

 * `MediaRelationship` was sending a custom function that called `handleError` when there was an error returned on the graphql `commitMutation`. However, this overwrote the function signature and did not send an `error` object to be consumed down the line. By switching to `handleError` itself, we send function with the correct signature expected by the rest of the code
 * In the process of resolving the above, Alex and I noticed we needed to handle cases where `getErrorMessage` in the `helpers.js` file is sent an empty `transactionOrError` object. We now handle it and default to the `fallbackMessage`.
 * `MediaSuggestionsComponent` could end up sending an `undefined` media object to `SmallMediaCard`, particularly in a case where someone bulk accepts all media suggestions. Suddenly there are no valid `relationshipItem`s but in some cases it still tries to render `SmallMediaCard`. -- this commit changes the code so that if a media item is missing we don't render a `SmallMediaCard` at all.

## Type of change

## Checklist

### DIFF
--- a/src/app/components/media/Similarity/MediaRelationship.js
+++ b/src/app/components/media/Similarity/MediaRelationship.js
@@ -113,9 +113,8 @@ const RelationshipMenu = ({
         browserHistory.push(`/${teamSlug}/media/${targetId}`);
       }
     },
-    () => {
-      handleError();
-    });
+    handleError,
+    );
   };
 
   const handleDelete = () => {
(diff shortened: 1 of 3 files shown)


dariusk:
Going to merge after tests pass since @amoedoamorim already approved this in our meeting

amoedoamorim (approved):

PR #1657, opened 2023-10-03
Bugfix/deploy urllib3 update

sonoransun:
## Description

This fix removed version pins for python packages to fix deployments.

## Type of change

## How has this been tested?

I verified a deployment in QA using this method.

## Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -21,6 +21,7 @@ build_qa:
     - docker push "$ECR_API_BASE_URL/qa/check/web:$CI_COMMIT_SHA"
   only:
     - develop
+    - bugfix/deploy-urllib3-update
 
 deploy_qa:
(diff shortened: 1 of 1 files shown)


dmou:
whoa, why are you removing them instead of updating them?

amoedoamorim (approved):

amoedoamorim:
> whoa, why are you removing them instead of updating them?

My understanding is we don't need those version-specific (if at all?) boto* 
We just need ecs-deploy and perhaps those were dependencies of that older version-pinned ecs-deploy?

dmou (changes requested):
blocking: removing version pins for a critical deployment step without a very good reason isn't my idea of good practice, ecs-deploy needs to be version pinned

dmou (approved):

PR #1666, opened 2023-10-09
Chat history view

dariusk:
Chat history is mostly done here. Some notes:

 * pagination is __implemented__ but not __enabled__ -- this is due to two pagination UX issues. One is that there is not currently a way to go to a specific message via the API without first grabbing the array of all messages, which defeats the point of pagination. Still, we could paginate by loading just the most recent ~10 messages and going back from there. But the problem here is the second issue: delivery confirmation messages are their own message -- it's possible we paginate and there's a boundary in between a message and its delivery confirmation, which is just weird. What do we do there? __For now I am not paginating while we address these issues__, but the components are in there.
 * I still need to figure out how to coerce the test bots to enter error states like delivery failure. These styles are not currently implemented. I believe the messages will show up, they just won't be styled as errors.

image

image

TODO: unit tests, BEM-like classnames for some remaining interactive elements

## Type of change

Does the data manipulation in `ChatHistory` make sense?

## Checklist


amoedoamorim:
image

This looks so good! Great work, Darius!
I see you have some todos and commented code. Are you planning to merge already?

src/app/components/annotations/TiplineHistoryButton.js
```diff
@@ -0,0 +1,50 @@
...
+    <>
+      <ButtonMain variant="outlined" theme="text" iconCenter={<QuestionAnswerIcon />} onClick={handleClick} className="int-tipline-history__button--open" />
+      <Modal
+        open={dialogOpen}
+        onClose={handleClose}
+      >
+        <TiplineHistory
+          uid={uid}
+          handleClose={handleClose}
+          title={`Chat with ${name}`}
```
amoedoamorim, line 34: Missing localization

amoedoamorim (approved):
This looks so good! All I have to nitpick about is one unlocalized string.

caiosba (commented):
Great work @dariusk ! I'm marking my review as "comment" because, when testing locally with WhatsApp, I noticed a few things that I don't see as blockers but that I think should be addressed (not necessarily in this PR). I pushed a fix for the only blocker I found (which was a crash). More details below:

Template messages are "encoded" messages, and they are currently displayed verbatim, for example:

Captura de tela de 2023-10-09 23-28-27

For those cases, it's enough to get only what's inside `fallback=[[]]`. It's safe to assume that anything that starts with `&((namespace` is a template message. That encoded format is enforced by Smooch.

Second, when a user votes that a search result is relevant, it's currently displayed this way:

Captura de tela de 2023-10-09 23-31-13

If the design doesn't define anything for that case, I suggest we just turn `search_result_is_relevant` (which is the ID associated with a button that was clicked) into something more readable. Or we can just hide `search_result_is_relevant`, since the button label is the thumbs up emoji, which is already being displayed.

I also received an exception when I was testing with WhatsApp Cloud API, related to templates too, but I pushed a commit to fix it (not the best fix, but enough for now I guess, probably requires a new design just to handle the template case):

Captura de tela de 2023-10-09 23-35-05

And just a few more questions:
- Is it intended that the scrollbar is not displayed?
- Is it possible to format text between asterisks (e.g., `*Title*`) as bold, like WhatsApp does?
- Can we linkify URLs?

dariusk:
Great, thanks @caiosba @amoedoamorim -- will address all these points today

dariusk:
@caiosba 

> Is it intended that the scrollbar is not displayed?

Yes. Otherwise the scrollbar displays over the curved div borders:

image

There are... complicated solutions to this that we can address in the future but for now the two solutions are either to remove the border-radius from the popover, or to make the scrollbar invisible. cc @brianfleming 

> Is it possible to format text between asterisks (e.g., `*Title*`) as bold, like WhatsApp does?

It's possible, but it's a huge can of worms. We can do a simple regex that is likely to fail on a bunch of use cases like:

```js
text.match(/\*+(?:([^*?$\n\r]+)\*+)?/)
```

but to do it for "real" you need a nontrivial parser. And then the question is, what does whatsapp support versus other systems? Like if `*title*` means `<strong>title</strong>` in WhatsApp, it might not mean the same thing in Telegram, at which point do we test what the source is and route the text through different parsers?

Basically I think it's too complex to tackle right now and maybe not even desirable. IMO the point of these markdown-light text formatting fallbacks is precisely so that they degrade to something human readable like `*Title*` and we _don't_ have to parse them

> Can we linkify URLs?

Yeah I'll see what I can do with our linkify function!

caiosba (approved):

PR #1669, opened 2023-10-10
Parse CAPI templates from WhatsApp

dariusk:
Fix to rendering WhatsApp CAPI templates:

BEFORE:

image

AFTER (something like this text):

image

..not perfect but a good start.


caiosba (approved):
Way better! Thanks Darius!

PR #1670, opened 2023-10-10
Handle smooch templates

dariusk:
Extracting the `fallback` for Smooch templates -- note that this is not in an object but rather rendered as a raw text message starting with the namespace prefix, so I check in the raw text parsing for the namespace prefix and then extract the `fallback` with a regex.

Also added more comments to make the CapiTemplate parser clearer.


amoedoamorim (approved):
:lgtm:

PR #1672, opened 2023-10-11
fix bug in smooch rendering

dariusk:
Fix the bug reported on CV2-3411

### DIFF
--- a/src/app/components/layout/ChatHistory.js
+++ b/src/app/components/layout/ChatHistory.js
@@ -44,7 +44,7 @@ const ChatHistory = ({
     let output = '';
 
     // Items from us have text directly in the payload, items from users have it in a messages sub-object
-    if (item.payload?.text) {
+    if (item.payload?.text && typeof item.payload.text.match === 'function') {
       // Smooch templates are raw text objects that start with the text `&((namespace` and look like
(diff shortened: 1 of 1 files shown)


src/app/components/layout/ChatHistory.js
```diff
@@ -44,7 +44,7 @@ const ChatHistory = ({
     let output = '';
 
     // Items from us have text directly in the payload, items from users have it in a messages sub-object
-    if (item.payload?.text) {
+    if (item.payload?.text && typeof item.payload.text.match === 'function') {
```
amoedoamorim, line 47: While this seems to do it, I'm wondering what the type of `item.payload.text` would be when it crashed. I'd think of fixing this slightly differently and have `&& typeof item.payload.text === 'string'` thinking that match is a given for String prototype.

 No big deal here, I'm just thinking that ultimately we want to test for the type of `text` instead that of `match`. Does that make sense?

amoedoamorim (approved):

src/app/components/layout/ChatHistory.js
caiosba: I'm going to merge this just to get the fix out since we need to verify the feature, but I had the same question as @amoedoamorim , @dariusk .

caiosba (approved):

src/app/components/layout/ChatHistory.js
dariusk: Makes sense. I've made a new PR, #1673

PR #1692, opened 2023-10-19
Create a sandbox route that crashes

dariusk:
This is to test situations where navigating to a new route causes a crash. The new route `/check/me/ui-sandbox/crash` points to an empty component that always crashes.

References: CV2-3874

## Type of change

### DIFF
--- a/src/app/components/SandboxCrash.js
+++ b/src/app/components/SandboxCrash.js
@@ -1,4 +1,3 @@
-/* eslint-disable relay/unused-fields */
 import React from 'react';
 import { QueryRenderer, graphql } from 'react-relay/compat';
 import Relay from 'react-relay/classic';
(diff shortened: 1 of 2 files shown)


src/app/components/SandboxCrash.js
```diff
@@ -0,0 +1,33 @@
...
+        }
+      }
+    `}
+    render={({ props }) => (
+      <SandboxCrashComponent admin={props?.me} />
+    )}
+  />
+);
+
+export default SandboxCrash;
```
amoedoamorim, line 32: We could even get rid of the query renderer to avoid compiling and generating unnecessary relay/graphql files and have, say,

const foo = {};

and render foo.bar.baz

what do you think?

dariusk: Good point.

amoedoamorim (approved):

PR #1710, opened 2023-10-30
Item page improvements to right panel tabs

brianfleming:
## Description

Meandering PR to keep improving the layout and style consistency of the item page, focusing on the right column tabs

- Ensure that the "empty" messages in the right panel are all visually the same
- Remove the counts from request tabs in the list, now that they are part of the tab name
- Fix some display logic issues around showing annotations when items are in the trash
- Switch various components to CDS versions
- Update the header for bulk actions on suggestions to help with upcoming work: [CV2-2947](https://meedan.atlassian.net/browse/CV2-2947)
- Fix layout issues with suggestions when screen gets narrow
- Fixed the tooltips on suggestions action buttons, they were not showing up correctly

## Type of change

- [x] Performance improvement and/or refactoring (non-breaking change that keeps existing functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Security mitigation or enhancement
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Automated test (add or update automated tests)

## How has this been tested?

manual testing

### BEFORE- Eliminated bug that does not correctly hide annotations when item is in the trash
Screenshot 2023-10-27 at 10 27 29 AM

### BEFORE - editing
Screenshot 2023-10-30 at 1 08 34 PM

### AFTER - Annotations of item in trash
Screenshot 2023-10-30 at 1 08 50 PM

### AFTER - Notes empty state
Screenshot 2023-11-06 at 12 40 57 PM

### AFTER - Populated Notes
Screenshot 2023-11-06 at 12 41 28 PM

### AFTER - Requests list
Screenshot 2023-11-06 at 12 41 16 PM

### AFTER - Suggestions empty state
Screenshot 2023-11-06 at 12 41 12 PM

### AFTER - Suggestions Bulk layout update
Screenshot 2023-11-06 at 3 51 38 PM

### AFTER - Source populated
Screenshot 2023-11-06 at 12 41 03 PM

### BEFORE -Suggestions Narrow
Screenshot 2023-11-06 at 12 42 00 PM

#### AFTER - Suggestions Narrow
Screenshot 2023-11-06 at 4 01 45 PM

## Checklist

- [x] I have performed a self-review of my own code
- [x] I've made sure my branch is runnable and given good testing steps in the PR description
- [x] I considered secure coding practices when writing this code. Any security concerns are noted above.
- [x] I have commented my code in hard-to-understand areas, if any
[...]

PR #1713, opened 2023-11-01
Update item page right panel tabs to include counts of items

brianfleming:
## Description

In order to make [this WIP PR](https://github.com/meedan/check-web/pull/1710) easier to review, I have broken off this funtionality into a separate PR.

- Remove the suggestions count from the media column. It is not a good interaction pattern to have a click update an area so far away
- As a result of ^^, Remove counterbutton component, it is not used anywhere else
- Add counts to the right panel tabs when the count is greater than 0

## Type of change

## How has this been tested?
manual testing

### Media Header after
Screenshot 2023-10-26 at 9 37 30 AM

### Counts added to tabs when > 0
Screenshot 2023-10-26 at 9 37 34 AM

Screenshot 2023-10-26 at 9 37 41 AM

## Checklist


src/app/components/media/MediaComponentRightPanel.js
```diff
@@ -37,8 +37,9 @@ const MediaComponentRightPanel = ({
             label={
               <FormattedMessage
                 id="mediaComponent.requests"
-                defaultMessage="Requests"
+                defaultMessage="{requests_count, plural, =0 {Requests} other {Requests [#]}}"
```
amoedoamorim, line 40: Please let's not use plurals for this and rather display count programatically. This formatting is not parsed by Transifex.

The formatting accepted by transifex is: `{total, plural, one {# tag} other {# tags}}` which seems to defeat the purpose of what you're trying to accomplish here.

amoedoamorim (changes requested):

src/app/components/media/MediaComponentRightPanel.js
brianfleming: I don't think I understand @amoedoamorim I grabbed my code from another part of the app, so I thought it was ok. Regardless, the only difference I see between my code and your formatting is:
- the use of [] around the number
- and the use of the =0 operator

Is one of those two things the issue?

brianfleming: this is what I used as the "starting" point:
https://github.com/meedan/check-web/blob/257c58eb1372e93f32e762e3d8c7c2c3ec7d72ef/src/app/components/user/UserTooltip.js#L125

src/app/components/media/MediaComponentRightPanel.js
```diff
@@ -83,12 +83,14 @@ const MediaComponentRightPanel = ({
         />
         <Tab
           label={
-            <FormattedMessage
-              id="mediaComponent.notes"
-              defaultMessage="{notes_count, plural, =0 {Notes} other {Notes [#]}}"
-              description="Label for the Notes tab, as in text notes"
-              values={{ notes_count: projectMedia.notes_count }}
-            />
+            <span>
```
brianfleming, line 86: @amoedoamorim here is a different approach, is this what you would suggest going with?

src/app/components/media/MediaComponentRightPanel.js
amoedoamorim: Yes, the issue seems to be with the =0 operator, it looks like transifex will only support `one` and `other` for english. Having `=0` in the string causes the pluralized string not be parsed correctly by Transifex and then making it more difficult for translators to write translations correctly.
This pattern has been used since the very beginning of Check localization but this issue was flagged recently by one of the translators and I ended forgetting to submit a fix. This PR just brought this issue back to my attention. 
 I'll send a PR for updating all instances with =0. Sorry for the confusion.

src/app/components/media/MediaComponentRightPanel.js
amoedoamorim: Yeah, this is better, thanks!

src/app/components/media/MediaComponentRightPanel.js
brianfleming: Nice, thanks for the info. I will update this PR to match the other pattern

brianfleming:
Ok @amoedoamorim thanks for the feedback, take another look, I updated the other areas of the PR with that change

PR #1721, opened 2023-11-07
Sentry quick fix: Switching to arrow-function

dariusk:
There was a bug in sentry where the `onFileError` event that occurs when someone attaches an invalid item (like wrong file type) to a Note would silently trigger an error about a missing `this`. It also meant that the error message was never correctly displayed on a note attachment file error.

Using an arrow function causes `this` to bind correctly, which means we actually update the state with the error message when an unsupported file is uploaded as seen here:

image

## Type of change

## How has this been tested?

Reproduced the bug locally, manually tested that the feature works now.

## Checklist

### DIFF
--- a/src/app/components/annotations/AddAnnotation.js
+++ b/src/app/components/annotations/AddAnnotation.js
@@ -389,6 +389,7 @@ class AddAnnotation extends Component {
           />
           {this.state.fileMode ? (
             <UploadFile
+              className="int-note-annotation__upload-file-drop-zone"
               type="file"
               value={this.state.file}
               onChange={this.onFileChange}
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -20072,7 +20072,7 @@
             "ms": {
               "version": "2.0.0",
               "resolved": "https://registry.npmjs.org/ms/-/ms-2.0.0.tgz",
-              "integrity": "sha512-Tpp60P6IUJDTuOq/5Z8cdskzJujfwqfOTkrwIwj7IRISpnkJnT6SyJ4PCPnGMoFjC9ddhal5KVIYtAt97ix05A=="
+              "integrity": "sha1-VgiurfwAvmwpAd9fmGF4jeDVl8g="
```
amoedoamorim, line 20075: Weaker sha alert!

dariusk: oops, I committed a bunch of files including localization stuff, reverting

amoedoamorim (approved):

PR #1730, opened 2023-11-15
Shared feed invitation accept interface

dariusk:
## App-wide changes

 - routes can now have a `splash` prop, which indicates to the `Home` component that they should be rendered without the context of any workspace (aka no left menu drawer). This works similarly to the existing `public` prop, which is for special-case routes that don't need login. Talking to design, we agreed this seemed like something they'll need for future features.
 - `Home` component checks for the `splash` prop and renders without nav drawer
 - minor (?) change to `ButtonMain` css that specifies a width for the icon container (otherwise it inherits parent widths and renders wrong)

## Feature-specific changes

 - New route, `:team/feed-invitation/:feedInvitationId`, which is the landing page that prompts a user to choose which workspace they want to associate with a feed invitation. Note: `:team` doesn't really do anything here at this point and could just as easily be `/check/me` or something.
 - `FeedInvitation` component, associated with the above route. This checks for several possible rendering states:
    - user is attempting to access an invitation they have no permission to (display `NotFound`)
    - user is attempting to access a nonexistent invitation (display `NotFound`) - user is admin of exactly one workspace (redirect to `FeedInvitationRespond`) - user is attempting to access an invitation they have already accepted, display info-level alert message and prevent action - none of the above happens and the component renders its default, which shows a list of workspaces the user is admin on. clicking a button on a workspace item takes them to `FeedInvitationRespond` for that workspace
 - New route, `:team/feed-invitation/:feedInvitationId/respond`, which prompts the user to accept or decline a feed invitation. Here, `:team` matters and is the workspace that will be associated to a feed, should it be accepted
 - `FeedInvitationRespond` component, associated with the above route. This checks for several possible rendering states:
    - user is attempting to access an invitation they have no permission to (display `NotFound`)
    - user is attempting to access a nonexistent invitation (display `NotFound`) - user is not admin of the current team (display error message) - user is attempting to access an invitation they have already accepted, display info-level alert message and prevent action - none of the above happens and the component renders its default, which prompts the user to accept/decline the invitation

References: CV2-3806

## Type of change

This wasn't specified in the design but seemed necessary, I added some info/error states as appropriate:

image

## Checklist

### DIFF
--- a/src/app/components/cds/buttons-checkboxes-chips/ButtonMain.module.css
+++ b/src/app/components/cds/buttons-checkboxes-chips/ButtonMain.module.css
@@ -28,6 +28,7 @@
     display: flex;
     /* stylelint-disable-next-line scale-unlimited/declaration-strict-value */
     font-size: 20px;
+    width: fit-content;
   }
 
   &.sizeSmall {
--- a/src/app/components/feed/FeedInvitation.js
+++ b/src/app/components/feed/FeedInvitation.js
@@ -11,15 +11,9 @@ import styles from './FeedInvitation.module.css';
 import ButtonMain from '../cds/buttons-checkboxes-chips/ButtonMain';
 import ScheduleSendIcon from '../../icons/schedule_send.svg';
 import DoneIcon from '../../icons/done.svg';
-import NotFound from '../NotFound';
 import { can } from '../Can';
 
 const FeedInvitationComponent = ({ routeParams, ...props }) => {
-  // `feed_invitation` is null when attempting to view a feed invitation that doesn't exist or that you don't have permission to view
-  if (!props.feed_invitation) {
-    return <NotFound />;
-  }
-
   // display an error if the user is not an admin on any workspaces
   const noAdminWorkspaces = props.me.teams?.edges.filter(team => can(team.node.permissions, 'update Team')).length === 0;
   const oneAdminWorkspace = props.me.teams?.edges.filter(team => can(team.node.permissions, 'update Team')).length === 1;
--- a/src/app/components/feed/FeedInvitation.js
+++ b/src/app/components/feed/FeedInvitation.js
@@ -20,7 +20,6 @@ const FeedInvitationComponent = ({ routeParams, ...props }) => {
     return <NotFound />;
   }
 
-  const feedInvitationId = parseInt(routeParams.feedInvitationId, 10);
   // display an error if the user is not an admin on any workspaces
   const noAdminWorkspaces = props.me.teams?.edges.filter(team => can(team.node.permissions, 'update Team')).length === 0;
   const oneAdminWorkspace = props.me.teams?.edges.filter(team => can(team.node.permissions, 'update Team')).length === 1;
@@ -29,10 +28,10 @@ const FeedInvitationComponent = ({ routeParams, ...props }) => {
   // redirect to the accept/decline page if the user only administers one workspace
   if (oneAdminWorkspace) {
     const team = props.me.teams.edges[0];
-    browserHistory.push(`/${team.node.slug}/feed-invitation/${feedInvitationId}/respond`);
+    browserHistory.push(`/${team.node.slug}/feed/${routeParams.feedId}/invitation`);
   }
 
-  const handleClick = team => browserHistory.push(`/${team.node.slug}/feed-invitation/${feedInvitationId}/respond`);
+  const handleClick = team => browserHistory.push(`/${team.node.slug}/feed/${routeParams.feedId}/invitation`);
 
   return (
     <div className={cx('feed-invitation-container', styles['feed-invitation-container'])}>
@@ -101,8 +100,7 @@ const FeedInvitationComponent = ({ routeParams, ...props }) => {
 
 FeedInvitationComponent.propTypes = {
   routeParams: PropTypes.shape({
-    team: PropTypes.string.isRequired,
-    feedInvitationId: PropTypes.string.isRequired,
+    feedId: PropTypes.string.isRequired,
   }).isRequired,
 };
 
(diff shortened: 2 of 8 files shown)


src/app/components/cds/buttons-checkboxes-chips/ButtonMain.module.css
```diff
@@ -28,6 +28,7 @@
     display: flex;
     /* stylelint-disable-next-line scale-unlimited/declaration-strict-value */
     font-size: 20px;
+    width: fit-content;
```
dariusk, line 31: @brianfleming I don't *think* this will affect anything currently in production, and the Sandbox looks good still. Basically this is just to prevent the width of the button icon from inheriting from a parent

brianfleming: Makes sense...I will click around and see if anything weird happens, but I highly doubt it

brianfleming (changes requested):
Two things @dariusk one of them was my mistake for not specifying on the spec:

1. We can make the whole "row" item clickable instead of just the "view invitation" part. I added that in the design just to show, that when you are hovering over the "row", if you click you will go to that invitation. I can see why it was not clear in the designs, but I think its better for the user to have that whole thing be a click target. Otherwise the styles look good
2. Can we sort the list of workspaces alphabetically by name? See screenshot below
Screenshot 2023-11-15 at 1 38 15 PM

src/app/components/feed/FeedInvitation.js
```diff
@@ -0,0 +1,153 @@
...
+import ScheduleSendIcon from '../../icons/schedule_send.svg';
+import DoneIcon from '../../icons/done.svg';
+import NotFound from '../NotFound';
+import { can } from '../Can';
+
+const FeedInvitationComponent = ({ routeParams, ...props }) => {
+  // `feed_invitation` is null when attempting to view a feed invitation that doesn't exist or that you don't have permission to view
+  if (!props.feed_invitation) {
+    return <NotFound />;
+  }
```
amoedoamorim, line 21: Nice! I'm just curious about why we need to manually check for this instead of having it handled by the API returning a 404 and being automatically redirected by the network layer, like it is in other cases such as attempting to load a team or media without the right permissions.

dariusk: Yeah, for whatever reason we don't get a 404 so this is just me handling it. I could open a bug for it?

brianfleming:
@dariusk I noticed another workflow thing that we can simplify.
If you are logged into a workspace that you have a pending invite for, when you click on it in the side menu, there is no need to go to the workspace selector page, you should just go straight to the invite page. See the video below. Let me know if I can explain better


https://github.com/meedan/check-web/assets/418001/652e6eb8-f24b-42a7-ac58-5b38ea7572d8

dariusk:
@brianfleming Sorting alphabetical is easy, I'll get that in. I'll see about making the whole row clickable -- eslint flagged some accessibility issues with that but maybe I can use an `aria-role` or something

dariusk:
@brianfleming Re clicking the side bar... I'm not sure we really want to do that? Ultimately the invitations really are at a person/email level and not a workspace level so I think that we do want to give someone every option. Like imagine if you've got multiple workspaces open in different tabs and you lost track of which workspace you were in in which tab and clicked the wrong one. Basically I think being explicit about it is better, and this is not a thing that happens very often so smoothing out the experience won't really save much time/friction for users in the end given the headaches that could be caused by not being explicit.

src/app/components/feed/FeedInvitation.js
amoedoamorim: Sounds like a good idea! I'd rather have this consistently handled by the network layer than having custom approaches in different places.

brianfleming:
> @brianfleming Re clicking the side bar... I'm not sure we really want to do that? Ultimately the invitations really are at a person/email level and not a workspace level so I think that we do want to give someone every option. Like imagine if you've got multiple workspaces open in different tabs and you lost track of which workspace you were in in which tab and clicked the wrong one. Basically I think being explicit about it is better, and this is not a thing that happens very often so smoothing out the experience won't really save much time/friction for users in the end given the headaches that could be caused by not being explicit.

Fair enough...I'm convinced.
related/unrelated question regarding this PR. Are you/we going to do the right side items (see figma screenshot) after @amoedoamorim other PR since he is working on those items as part of the collaborator edit screen? or do you plan on doing that in this PR? Just want to make sure I know what to review

Screenshot 2023-11-15 at 1 53 25 PM

dariusk:
@brianfleming The sidebar elements are currently not reusable from Alex's work -- I think we need another ticket to generalize those components. Otherwise I'm making a fair amount of duplicate code.

dariusk:
@brianfleming Just pushed workspace alphabetical sorting and made the row clickable in eab4fa892b1e91b1590b363cde8eb8c4e0cd0129

brianfleming:
> @brianfleming The sidebar elements are currently not reusable from Alex's work -- I think we need another ticket to generalize those components. Otherwise I'm making a fair amount of duplicate code.

ok, those are important items for the design, because the organization accepting the invitation needs to be able to see the "metadata" of the shared feed so we are not hiding any info from them.

dariusk:
@brianfleming I can do a big old copy/paste for now, I actually have it in a temp branch like that...

amoedoamorim:
I can extract it as a component

brianfleming:
> I can extract it as a component

That would be good @amoedoamorim 
I didn't mean to "just" highlight the meta data as well, the collaborator list is important too

dariusk:
I'm going to merge this so we can test the core, non-sidebar features ASAP. I'll open a new pull request for the remaining work, which is integrating the sidebar components.

PR #1733, opened 2023-11-15
L10n/update nov 15

amoedoamorim:
## Description

Just realized that some locales were not being correctly generated! :o


caiosba (approved):
Oh, I see, it includes `id`, `ta` and `ur` that were missing.

PR #1740, opened 2023-11-23
Fix checking to display "more actions" button

danielevalverde:
## Description

- Merge the permissions of feed and feedTeam
- verify whats is the permissions required if the user is from an organization invited or a feed owner

Reference: CV2-3805-leave-shared-feeds
## Type of change

## Checklist

### DIFF
--- a/src/app/components/feed/SaveFeed.js
+++ b/src/app/components/feed/SaveFeed.js
@@ -486,11 +486,11 @@ const SaveFeed = (props) => {
           />
           { feed.id ?
             <FeedActions
+              feedTeam={{ ...feedTeam, permissions: feedTeam.permissions }}
               disableSaveButton={disableSaveButton}
               saving={saving}
               handleDelete={handleDelete}
               handleLeaveFeed={handleLeaveFeed}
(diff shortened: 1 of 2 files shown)


caiosba (approved):
Thanks Dani! It reflects everything we talked about.

src/app/components/feed/SaveFeed.js
```diff
@@ -486,11 +486,11 @@ const SaveFeed = (props) => {
           />
           { feed.id ?
             <FeedActions
+              feedTeam={{ ...feedTeam, permissions: feedTeam.permissions }}
```
amoedoamorim, line 489: Is this necessary? Isn't permissions already part of the `feedTeam` prop/object?

caiosba: I think this was to make the linter happy, otherwise it would say that the field was requested but not used.

amoedoamorim: I see that it was added to `fragment FeedActions_feedTeam on FeedTeam` so it shouldn't be necessary here

PR #1778, opened 2023-12-21
Item Card refactor and bugfix

dariusk:
As preliminary work for CV2-3991, I've broken the `Card` component into three subcomponents: one for displaying a date, one for displaying the collapsible title/description, and one for displaying the item rating. The idea here is to eventually turn `Card` into just a thing that you hover over and it gets a highlight and also can double as a clickable link card if you pass it a url, but otherwise to be content-agnostic. Then we can have the layout for fact checks defined in `FactCheckCard`, the layout for shared feed item cards in a `SharedFeedItemCard`, etc etc. So ultimately our display cards will look like:

 * an overall layout logic component like `SharedFeedItemCard`
   * `Card`: basically just the rounded box container with hover and link behavior
     * a bunch of components we can pick and choose that actually do things in the layout like `ItemRating`

`Card` itself isn't refactored yet but we'll get there.

This also contains some bug fixes to behavior for the collapsible description fields: for example, it fixes an issue where long titles remained collapsed when a user pressed the expand button. It also fixes an issue where text would jump around on hover. There are also minor layout changes to make things more consistent with the CDS design (making the rating icon 16px instead of 14px, fixing some font settings, using `gap` instead of a space character to define spacing, etc).

@brianfleming In particular, if you could run it and look at the sandbox cards. I changed the hover behavior so that the show/hide for the expand/collapse button no longer causes the text box size to change; I also fixed an issue where long titles were never expanded and tweaked some layout stuff. Basically let me know if there's any tweaks, I tried to get it as close to the design system as possible.

image

References: CV2-3991

## Type of change

## Checklist


src/app/components/cds/media-cards/Card.js
```diff
@@ -78,51 +44,14 @@ const Card = ({
...
-            <h6 className={`typography-button ${styles.cardTitle}`}>{title}</h6>
-            { description ?
-              <p className="typography-body2">
-                <span className={`description-text ${styles.cardDescription} ${isCollapsed ? styles.cardDescriptionCollapse : ''}`} ref={descriptionRef}>
-                  {description}
-                </span>
-                { factCheckUrl ?
-                  <span className={styles.factCheckLink}>
-                    <FactCheckIcon />
-                    {' '}
```
dariusk, line 91: I wanna get rid of stuff like this, we shouldn't be spacing with text!

src/app/components/cds/media-cards/ItemDescription.js
```diff
@@ -0,0 +1,90 @@
...
+import UnfoldLessIcon from '../../../icons/unfold_less.svg';
+import UnfoldMoreIcon from '../../../icons/unfold_more.svg';
+import FactCheckIcon from '../../../icons/fact_check.svg';
+import styles from './Card.module.css';
+
+const ItemDescription = ({
+  description,
+  title,
+  showCollapseButton,
+  factCheckUrl,
```
brianfleming, line 13: just more of a note, we will get to it eventually, but there will be the possibility of having other links including/instead of the fact check

src/app/components/cds/media-cards/Card.js
brianfleming: agreed...css ftw

brianfleming (approved):
@dariusk I think this is good progress, and doesn't break but only improve existing functionality.
The "tricky" parts of this come later with the other permutations as outlined in the prototyping area at the top that I see you started to integrate with, which is great.
So I don't see anything here that is going down the wrong path, but we will see as things move forward.

FWIW, the existing card that this PR improves on, the one that represents a card:
- in a shared feed
- the shared feed is only sharing fact-checks, no media
Would be replicated by this set of switches:
Screenshot 2023-12-21 at 1 04 12 PM

dariusk:
@brianfleming Yeah so ultimately instead of having a component that is sometimes a shared feed item and sometimes a workspace item, the idea here is to have a `SharedFeedItemCard` and a `WorkspaceItemCard` (for example) that have their own layout and don't need to conditionally lay out. So the sandbox would look like 2 different cards for those things because that's what they are. Does that make sense? The shared code between them would be subcomponents like `ItemDescription` and `ItemDate` etc.

brianfleming:
> @brianfleming Yeah so ultimately instead of having a component that is sometimes a shared feed item and sometimes a workspace item, the idea here is to have a `SharedFeedItemCard` and a `WorkspaceItemCard` (for example) that have their own layout and don't need to conditionally lay out. So the sandbox would look like 2 different cards for those things because that's what they are. Does that make sense? The shared code between them would be subcomponents like `ItemDescription` and `ItemDate` etc.

@dariusk sure, that is fine with me if that is the tech decision. Just keep in mind that both types of cards have permutations. So its not like one is less dynamic than the other. They both share the same basic sub components

PR #1785, opened 2023-12-22
Refactor Card and FactCheckCard

dariusk:
I'm considering `Card` now to be an empty rectangle container that can be passed a `cardUrl` and a `footer`, and highlights when you mouse over it. The card manages an `isHovered` state that is provided to its `children` via `Context` as needed. So in this case `FactCheckCard` renders a `Card` with some children in the correct layout. The `ItemDescription` is wrapped with the `CardHoverContext.Consumer`, because `ItemDescription` needs to know that `Card` (not `FactCheckCard`!) is being hovered. In other cases I would do this by passing state from Card to its children, but since we have a generic `children` object I'm doing it this way instead. Seems to work well.

References: CV2-3991

## Type of change


caiosba (approved):
Code looks good! I like that now `<Card />` is way simpler. I tested it locally with the fact-checks cards and it works well (I had to push a fix though, which I talked to @dariusk before). The only issue I noticed is that ellipsis are now missing for long URLs:

Captura de tela de 2023-12-22 17-54-47

But I'm approving for now in order to unblock, but please fix it in the next PR, @dariusk.

dariusk:
@caiosba I made a bunch of updates -- please see CV2-3991 comments for details but basically I have added:

 * sorted list of platforms that requests came from for a `SharedItemCard` (called `ItemChannels`)
 * helper function for compact number formatting (and longer number formatting)
 * conditional media icon when the media count is 1
 * tooltips for workspaces, dates, and counts - these are pretty complex and I explain more in CV2-3991

src/app/components/cds/media-cards/ItemChannels.js
```diff
@@ -0,0 +1,104 @@
...
+      case 'WHATSAPP': return <WhatsAppIcon style={{ color: 'var(--whatsappGreen)' }} />;
+      case 'MESSENGER': return <FacebookIcon style={{ color: 'var(--facebookBlue)' }} />;
+      case 'TWITTER': return <TwitterIcon style={{ color: 'var(--xBlack)' }} />;
+      case 'TELEGRAM': return <TelegramIcon style={{ color: 'var(--telegramBlue)' }} />;
+      case 'VIBER': return <ViberIcon style={{ color: 'var(--viberPurple)' }} />;
+      case 'LINE': return <LineIcon style={{ color: 'var(--lineGreen)' }} />;
+      case 'INSTAGRAM': return <InstagramIcon style={{ color: 'var(--instagramPink)' }} />;
+      default: return null;
+      }
+    }
```
caiosba, line 55: Question/suggestion: What if we move this mapping to https://github.com/meedan/check-web/blob/develop/src/app/CheckChannels.js? We do something similar in https://github.com/meedan/check-web/blob/develop/src/app/CheckFeedLicenses.js.

It can be out of scope for this ticket too - maybe we can just create a small refactoring ticket to abstract that to https://github.com/meedan/check-web/blob/develop/src/app/CheckChannels.js and replace it in other places it's done (like the channels filter).

src/app/components/cds/media-cards/ItemChannels.js
```diff
@@ -0,0 +1,104 @@
...
+
+ItemChannels.defaultProps = {
+  className: null,
+  sortMainFirst: false,
+};
+
+ItemChannels.propTypes = {
+  className: PropTypes.string,
+  sortMainFirst: PropTypes.bool,
+  channels: PropTypes.exact({
```
caiosba, line 98: I like `exact` instead of `shapeOf`!

src/app/components/search/SearchResultsCards/SharedItemCard.js
```diff
@@ -0,0 +1,193 @@
...
+  factCheckUrl: PropTypes.string,
+  date: PropTypes.number, // Timestamp
+  lastRequestDate: PropTypes.number, // Timestamp
+  mediaCount: PropTypes.number,
+  requestsCount: PropTypes.number,
+  intl: intlShape.isRequired,
+  channels: PropTypes.exact({
+    main: PropTypes.number,
+    others: PropTypes.arrayOf(PropTypes.number),
+  }),
```
caiosba, line 190: Missing `propTypes`:

- `mediaThumbnail`
- `mediaType`
- `workspaces`

caiosba (changes requested):
@dariusk , looks good! I added a couple of comments, and locally I noticed that a truncation bug (I think it's related to the URL) still happens (the URL should truncate with ellipsis):

pr-1785-bug-1

src/app/components/cds/media-cards/ItemChannels.js
dariusk: Good idea. Filed CV2-4181 for the refactor.

dariusk:
> @dariusk , looks good! I added a couple of comments, and locally I noticed that a truncation bug (I think it's related to the URL) still happens (the URL should truncate with ellipsis):
> 
> pr-1785-bug-1

@caiosba It looks like this is on some part of the Check app -- this component is not anywhere except for `Sandbox` right now, and it appears to be truncating correctly there:

image

You'll notice in the above screenshot that when the width is low, 1) the ellipsis truncation happens, and 2) the elements stack in a responsive flex flow.

caiosba:
Thanks @dariusk , can you please try with a long URL? I see it working with long titles and summaries in the fact-check card too.

dariusk:
@caiosba Long url does break it, looking into it.

dariusk:
I fixed the bug with displaying fact check urls that are very long, and I merged @Caio Almeida's work for `CheckFeedDataPoints`, adding the logic for the different display states for `SharedFeedItemCard` based on the data composition of a shared feed:

image

I added three switches to the sandbox - two of them to toggle what kind of data points are in a shared feed, and one to toggle the number of fact checks between 1 and 12345.

src/app/components/Sandbox.js
```diff
@@ -486,6 +504,42 @@ const SandboxComponent = ({ admin }) => {
...
+              </li>
+              <li>
+                <SwitchComponent
+                  label="Fact Check Count - More than 1"
+                  labelPlacement="top"
+                  checked={listItemFactCheckCount > 1}
+                  onChange={() => {
+                    if (listItemFactCheckCount === 1) {
+                      // eslint-disable-next-line
+                      console.log('~~~SET TO 12345');
```
caiosba, line 533: Just a reminder to delete this before merging

src/app/components/Sandbox.js
```diff
@@ -486,6 +504,42 @@ const SandboxComponent = ({ admin }) => {
...
+                      setListItemFactCheckCount(12345);
+                    } else {
+                      // eslint-disable-next-line
+                      console.log('~~~SET TO 1');
```
caiosba, line 537: Just a reminder to delete this before merging

caiosba (approved):
Thanks @dariusk , looks good now!

PR #1786, opened 2024-01-02
New feature: Custom item title

caiosba:
## Description

This PR adds a new feature: custom item title. On the item page, the header now shows a component where the title for an item can be customized. The user is able to type a custom title or choose between claim title, fact-check title and pinned media ID. See referenced ticket for details on designs and specs.

Reference: CV2-3986.

## Type of change

## How has this been tested?

I implemented unit tests for the new component `<ItemTitle />`. When testing manually, it's also important to be sure that the claim title and fact-check title are updated in this new component when they are updated from their respective components.

In terms of coding, most of the work here happens in a new component `<ItemTitle />`. Very little was changed in existing code.

I used states and not only props on this new component because there were situations where the component was not re-rendering (I think because props were changed from inside and so not propagating). I thought it was simpler to just be sure that the component had its own state.

I noticed that in recent components, like the feed quick filters, we're still using Material UI menu components, so that's what I used here as well.

I have some other design questions for @brianfleming:

- In the design, there are copies "Pinned Media Title" and "Pinned Media ID": Which of them is the right copy I should use?
- I'm using the size "large" for the pencil icon, as per the design, but it looks a bit too large, please review
- Design for permissions: I didn't find in the designs, so, for now, what I did is that the button is disabled if the user has no permission to update an item or if the item is in the trash... please let me know if permissions should be handled differently
- Design for error states: I didn't find it in the designs, so, for now, I display an error on the `<TextField />` component (using the `error` and `helpContent` props... please let me know if it should be something else

## Checklist

- [x] If my components involve user interaction - specifically button, text fields, or other inputs - I have added a [BEM-like class name](https://meedan.atlassian.net/wiki/spaces/ENG/pages/1327628289/Naming+conventions+for+interactive+elements) to the element that is interacted with


src/app/components/media/ItemTitle.js
```diff
@@ -0,0 +1,291 @@
...
+        }
+      />
+    </MenuItem>
+  );
+
+  return (
+    <div className={styles.itemTitle}>
+      <TextField
+        key={`${titleField}-${title}-${saving}`}
+        fullWidth
```
brianfleming, line 143: `fullWidth` is not used in our custom CDS component, it can be removed

src/app/components/media/ItemTitle.js
```diff
@@ -0,0 +1,291 @@
...
+        className={cx(styles.itemTitleInputField, 'int-item-title__textfield--title')}
+        defaultValue={title}
+        margin="normal"
```
brianfleming, line 146: `margin` is not used in our CDS component, it can be removed

src/app/components/media/ItemTitle.module.css
```diff
@@ -0,0 +1,48 @@
...
+
+.itemTitleInputField {
+  width: 100%;
+}
+
+.itemTitleMenuHeader {
+  color: var(--textSecondary);
+}
+
+.itemTitleOptionIcon {
```
brianfleming, line 15: more of a suggestion than a necessary change, but you can avoid using the `!important` in these css statements by using the css loader global scope, such as `:global(.MuiListItemText-root) {...`

what that does is, "breaks" out of the css modules rewriting concept and just uses the plain classname so you can target things like material components, etc.

brianfleming (changes requested):
Looking good @caiosba . Its great to come back from a break to new features to review, very exciting.

I left some inline comments, but also a few other things:

- [x] if you change `.media-actions` to be `flex: 0 1 auto;` then the new custom title area can expand and fill the remaining width of the title area, giving the user a larger textfield depending on their screen size
- [x] When using the custom title, we need to clean up the interaction a bit, since its misleading. Here is what I am seeing:
  - Go to an item with an auto selected title
  - Change to custom title
  - Notice the textfield is blank
  - do nothing, and just close the tab
  - Go to that same item in a new fresh tab
  - You will see that the custom title was actually set to the what the previous auto title was
 
So this is a bit confusing for the user, since it did not appear to them that they set the title.

**Suggestions:**
- When switching to custom title, ensure that the text field is filled with the content from the auto title (probably best idea)
**OR**
- Ensure the textfield has placeholder text instructing the user what to do, and do not actually set the custom title until the user changes the content of the textfield

caiosba:
Thanks @brianfleming !

I think this is the best suggestion: "When switching to custom title, ensure that the text field is filled with the content from the auto title (probably best idea)".
Because this is actually what I'm currently doing on the backend side... since choosing the title field is an atomic operation, when "custom title" is chosen, I need to set a value for the custom title, otherwise it will fail the backend validation. So, I'm already setting it to the auto title, I'll be sure to reflect that value in the text field.

I'm going to address the other comments too.

When you have a chance, please also answer the few questions I left in the PR:

- In the design, there are copies "Pinned Media Title" and "Pinned Media ID": Which of them is the right copy I should use?
- I'm using the size "large" for the pencil icon, as per the design, but it looks a bit too large, please review
- Design for permissions: I didn't find in the designs, so, for now, what I did is that the button is disabled if the user has no permission to update an item or if the item is in the trash... please let me know if permissions should be handled differently
- Design for error states: I didn't find it in the designs, so, for now, I display an error on the `<TextField />` component (using the `error` and `helpContent` props... please let me know if it should be something else

brianfleming:
> Thanks @brianfleming !
> 
> I think this is the best suggestion: "When switching to custom title, ensure that the text field is filled with the content from the auto title (probably best idea)". Because this is actually what I'm currently doing on the backend side... since choosing the title field is an atomic operation, when "custom title" is chosen, I need to set a value for the custom title, otherwise it will fail the backend validation. So, I'm already setting it to the auto title, I'll be sure to reflect that value in the text field.
> 
> I'm going to address the other comments too.
> 
> When you have a chance, please also answer the few questions I left in the PR:
> 
> * In the design, there are copies "Pinned Media Title" and "Pinned Media ID": Which of them is the right copy I should use?
> * I'm using the size "large" for the pencil icon, as per the design, but it looks a bit too large, please review
> * Design for permissions: I didn't find in the designs, so, for now, what I did is that the button is disabled if the user has no permission to update an item or if the item is in the trash... please let me know if permissions should be handled differently
> * Design for error states: I didn't find it in the designs, so, for now, I display an error on the `<TextField />` component (using the `error` and `helpContent` props... please let me know if it should be something else

Apologies @caiosba I meant to respond to those questions during my review...then...well just didn't! 😖
So here goes:

- Lets go with `Pinned Media Title`
  - Just for a little info sharing, this got changed during the design process after discussing with Pierre how we should eventually improve the media auto generated ID, so eventually its more like a title. Just an FYI for any readers of this PR
- I agree about the button size. Can you change it to match what we now have for select and textfield clear buttons. Check out the sandbox and you will see this:
Screenshot 2024-01-02 at 10 34 39 AM

- For permissions, forgot about this case. I think instead of disabling lets remove the button completely. That was for things like trash, the "Restore from Trash" button kinda takes the place of the action for this item.
- Errors: perfect, lets go with what you implemented

src/app/components/media/ItemTitle.js
```diff
@@ -0,0 +1,291 @@
...
+      }),
+    }),
+  }).isRequired,
+};
+
+// For unit tests
+// eslint-disable-next-line import/no-unused-modules
+export { ItemTitle };
+
+export default createFragmentContainer(ItemTitle, {
```
dariusk, line 273: Ideally I would like to see this component as a `QueryRenderer` passed a `dbid` for the `project_media` query, like we do in `MediaTags`:

https://github.com/meedan/check-web/blob/develop/src/app/components/media/MediaTags.js

Especially since we don't really need an `ItemTitle_projectMedia` fragment, right?

In `MediaTags` we pass in `projectMediaId` and then every piece derives from the query based on that id. So you can see in `MediaActionsBar` right now that the components are called like:

```jsx
          <ItemTitle projectMedia={this.props.media} />
          <MediaTags projectMediaId={this.props.media?.dbid} />
```

What I'd like to see is

```jsx
          <ItemTitle projectMediaId={this.props.media?.dbid} />
          <MediaTags projectMediaId={this.props.media?.dbid} />
```

Basically it makes it clear that all we really need for `ItemTitle` to work is the dbid, and that it's all self contained from a query, rather than receiving a huge projectmedia object.

dariusk (changes requested):

caiosba:
Thanks @brianfleming and @dariusk for the review - I addressed all the comments, kindly re-review please.

src/app/components/media/ItemTitle.js
caiosba: I shared this with @dariusk privately, but just to document it here as well:

There is an issue that happens on Relay when there are two `QueryRenderer`'s for the same route at the same level.

For example, for the suggestion below:

```jsx
  <ItemTitle projectMediaId={this.props.media?.dbid} />
  <MediaTags projectMediaId={this.props.media?.dbid} />
```

So, if `<ItemTitle />` has a `<QueryRenderer />` that asks for `project_media { title }` and `<MediaTags />` has a `<QueryRenderer />` that asks for `project_media { tags }`, this is what happens:

- If `<ItemTitle />` comes first, then `project_media.title` will be present but `project_media.tags` will be null, so `<MediaTags />` crashes
- If `<MediaTags />` comes first, then `project_media.tags` will be present but `project_media.title` will be null, so `<ItemTitle />` crashes

So, the `project_media` returned by one of the `QueryRenderer`'s fully replaces (in the Relay store) the one returned by the other `QueryRenderer` instead of merging them... it's as if Relay "forces" you to avoid more than one `<QueryRenderer />` for the same page.

I worked around that by differentiating the `variables` (the `QueryRenderer` `prop`) names between these components.

It also made me realize that it's important for us to document in our frontend good practices document when to use a fragment container or a query renderer, and other details like this issue I just described.

brianfleming (changes requested):
Thanks for those changes @caiosba looking good!
I noticed 2 things we should address, one important, one nit.
- [x] **nit:** can you add a bit of margin or padding to the top of the status container so it pushes down a bit a lines up with the button change we discussed? so it ends up like this:
Screenshot 2024-01-03 at 9 05 58 AM

- [x] **Important:** If there is no media for an item we should disable being able to set the title as the pinned media, since it results in an empty title. Seen here:

Screenshot 2024-01-03 at 9 08 50 AM

brianfleming (approved):
Nice work @caiosba !
Happy to approve with the latest changes

name

PR #1808, opened 2024-01-23
Workspace Item Card

dariusk:
This is a PR for the implementation of the `WorkspaceItemCard` -- it is currently running in the Trash section, and seems to work just fine in that context. @amoedoamorim if I could get a code check on this, though for some reason the PR includes some code Caio wrote for a different branch, not sure why since I merged it from `develop`. Anyway, the main files to check are `WorkspaceItemCard` and `SearchResults` for the trash implementation.

* already done
  * Media Preview / `ItemThumbnail`
  * `ItemChannels`
  * `ItemDescription`
  * `ItemDate`
* to do
  - [x] create `WorkspaceItemCard` that uses `Card`
    - [x] fact-check
      * `rating` string ("false", "undetermined", etc)
      * `ratingColor` string for corresponding color
    - [x] `published` boolean prop, controls the check mark next to a fact check, ignored if no fact check
    - [x] `unread` boolean prop to control the unread indicator
    - [x] `checked` boolean and `onChangeChecked` function props
    - same as on `SharedItemCard`
      - [x] `mediaThumbnail`
      - [x] `title`
      - [x] `description`
      - [x] `date`
      - [x] `lastRequestDate`
      - [x] `factCheckUrl`
      - [x] `mediaCount`
      - [x] `mediaType`
      - [x] `requestsCount`
      - [x] `channels`
    * NO `workspaces`, `dataPoints` needed
  * separate out the following subcomponents (these are implemented in `SharedItemCard`, just need to be moved into their own files)
    - [x] `LastRequestDate`
    - [x] `MediaCount`
    - [x] `RequestsCount`
  - [x] create `SuggestionsCount` subcomponent similar to the above
  - [x] unread status indicator, won't be a subcomponent but rather a piece of `WorkspaceItemCard`
  - [x] bulk selector checkbox, won't be a subcomponent but rather a piece of `WorkspaceItemCard`, should call `onChangeChecked` when the checkbox is selected or deselected
  - [x] add "empty state" to `ItemThumbail` with `empty_media.svg`

References: CV2-3991

## Type of change

## Checklist


dariusk:
image

* There is now a `WorkspaceItemCard` that uses `Card`
* The `WorkspaceItemCard` currently renders `ItemDescription` and `ItemThumbnail`
* I've added an unread indicator controlled by an `isUnread` prop on the component.
* I've added an "empty state" to `ItemThumbnail`. @caiosba Please take a close look at `ItemThumbnail`'s `propTypes` -- I implemented some custom prop type functions. Basically, it means that if we specify a `type` then it expects all fields to be required, but if `type` is empty, then no fields are required and it returns the new "empty state" icon seen in the screenshot above. I'm not sure we use custom proptypes anywhere else in the code so I wanted to flag that.
* Also I fixed an issue with `SharedItemCard` (in its own commit), which now accepts a `rating` and `ratingColor` prop rather than having a red `False` hard-coded

I've checked off all to-do items in the PR description to match this work. It looks like we are about 33% done.

dariusk:
Re: `ItemThumbnail`, we could also have an `isEmpty` prop that we set to `true` instead of testing if `type` is falsy. It would be more explicit, but also it would be another prop. I see advantages to both implementations.

src/app/components/cds/media-cards/LastRequestDate.js
```diff
@@ -0,0 +1,46 @@
...
+  <FormattedMessage id="sharedItemCard.lastRequested" defaultMessage="Last Requested" description="This appears as a label before a date with a colon between them, like 'Last Requested: May 5, 2023'.">
+    { lastRequestDateLabel => (
+      <Tooltip
+        arrow
+        title={(
+          <>
+            <span>{lastRequestDateLabel}:</span>
+            <ul>
+              <li>{Intl.DateTimeFormat('en-US', { year: 'numeric', month: 'long', day: 'numeric' }).format(lastRequestDate)}</li>
+              <li>{Intl.DateTimeFormat('en-US', { hour: 'numeric', minute: 'numeric' }).format(lastRequestDate)}</li>
```
amoedoamorim, line 20: Why is 'en-US' locale hardcoded? Isn't this supposed to follow Check UI locale?

src/app/components/cds/media-cards/MediaCount.js
```diff
@@ -0,0 +1,47 @@
...
+const MediaCount = ({
+  intl,
+  mediaCount,
+  mediaType,
+}) => (
+  <FormattedMessage id="sharedItemCard.medias" defaultMessage="Medias" description="This appears as a label next to a number, like '1,234 Medias'. It should indicate to the user that whatever number they are viewing represents the number of medias attached to an item .">
+    { mediasLabel => (
+      <Tooltip
+        arrow
+        title={`${getSeparatedNumber(intl.locale, mediaCount)} ${mediasLabel}`}
```
amoedoamorim, line 19: I think it would be more appropriate to make the localizable string pluralized to account for 0, 1..N medias and also for better translation. We can't be sure that all languages will be translated number then "medias" in the same order.

src/app/components/cds/media-cards/LastRequestDate.js
```diff
@@ -0,0 +1,48 @@
...
+}) => (
+  <FormattedMessage id="sharedItemCard.lastRequested" defaultMessage="Last Requested" description="This appears as a label before a date with a colon between them, like 'Last Requested: May 5, 2023'.">
+    { lastRequestDateLabel => (
+      <Tooltip
+        arrow
+        title={(
+          <>
+            <span>{lastRequestDateLabel}:</span>
+            <ul>
+              <li>{Intl.DateTimeFormat(intl.locale, { year: 'numeric', month: 'long', day: 'numeric' }).format(lastRequestDate)}</li>
```
dariusk, line 20: @amoedoamorim Fixed this to `intl.locale`, thanks!

src/app/components/cds/media-cards/MediaCount.js
```diff
@@ -0,0 +1,53 @@
...
+import { getCompactNumber } from '../../../helpers';
+
+const MediaCount = ({
+  intl,
+  mediaCount,
+  mediaType,
+}) => (
+  <FormattedMessage
+    id="sharedItemCard.medias"
+    // {mediaCount, number} tells react-intl to format the number as Intl.NumberFormat(locale, {})
```
dariusk, line 17: @amoedoamorim Thanks for the review. I totally changed how the labels work here so instead of a separately localized number and a separately pluralized string, we are using the internal react-intl NumberFormat API and doing a fully localized string+number together, embedded in the `plural` directive.

src/app/components/search/SearchResultsCards/SharedItemCardFooter.js
```diff
@@ -1,116 +1,50 @@
...
 import ButtonMain from '../../cds/buttons-checkboxes-chips/ButtonMain';
-import Tooltip from '../../cds/alerts-and-prompts/Tooltip';
-import MediaTypeDisplayIcon from '../../media/MediaTypeDisplayIcon';
-import MediaIcon from '../../../icons/perm_media.svg';
-import CalendarMonthIcon from '../../../icons/calendar_month.svg';
-import RequestsIcon from '../../../icons/question_answer.svg';
 import OpenInNewIcon from '../../../icons/open_in_new.svg';
 import ItemChannels from '../../cds/media-cards/ItemChannels';
-import { getCompactNumber, getSeparatedNumber } from '../../../helpers';
+import SuggestionsCount from '../../cds/media-cards/SuggestionsCount';
```
dariusk, line 8: Flagging for @caiosba that I refactored `SharedItemCardFooter`. Specifically the individual pieces are componentized, and for the counts we want to check `variable !== null` to see if we render, since `0` is a valid case here (we want to show "0 Requests" if we pass in `requestsCount={0}`

caiosba: Sounds good, thanks @dariusk! So let's please add smoke-testing feeds to the testing notes just to be sure that the refactoring didn't include any regressions.

src/app/components/search/SearchResultsTable/ItemThumbnail.js
```diff
@@ -65,11 +75,33 @@ const ItemThumbnail = ({
...
+
 ItemThumbnail.propTypes = {
-  url: PropTypes.string.isRequired,
-  type: PropTypes.string.isRequired,
-  picture: PropTypes.string.isRequired,
-  maskContent: PropTypes.bool.isRequired,
+  type: PropTypes.string,
+  url: requiredStringIfTypeNotEmpty,
+  picture: requiredStringIfTypeNotEmpty,
+  maskContent: requiredBoolIfTypeNotEmpty,
```
amoedoamorim, line 104: Awesome! :exploding_head:

amoedoamorim (approved):

src/app/components/search/SearchResultsCards/SharedItemCardFooter.js
dariusk: Ok! Added to the test plan

PR #1844, opened 2024-03-06
Apply some UI tweaks for shared feed media clusters list

danielevalverde:
## Description

- Change page size from 10 to 50, so it becomes consistent with other lists

- Default sorting: Number of requests, descending

- Bug: Some cards don’t have a title - we need to avoid empty titles

- Pluralization must be fixed for some card components:

- Add workspace name tooltip to the Fact-Check only card, already exists on cluster card

- Add a tooltip for the media thumbnail to indicate type of media

Reference: CV2-4354

## Type of change

## Checklist

### DIFF
--- a/src/app/components/search/SearchResultsTable/ItemThumbnail.js
+++ b/src/app/components/search/SearchResultsTable/ItemThumbnail.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import VisibilityOffIcon from '../../../icons/visibility_off.svg';
 import styles from './ItemThumbnail.module.css';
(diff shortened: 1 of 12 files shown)


src/app/components/search/SearchResultsTable/ItemThumbnail.js
```diff
@@ -33,7 +34,9 @@ const ItemThumbnail = ({
     return (
       <div className={`${styles.thumbnail} ${styles.container}`}>
         <div className={styles.iconContainer}>
-          <MediaTypeDisplayIcon mediaType={mediaType} className={styles.mediaIcon} fontSize="var(--iconSizeDefault)" />
+          <Tooltip title={mediaType} arrow>
```
caiosba, line 37: I was not able to make this work, Dani... but in any case, we should translate this string, not show the `mediaType` in English.

caiosba (changes requested):
Thanks Dani! I tested:

- [ ] Default sort / order
- [ ] Pluralization for number of requests and number of media items
- [ ] Page size
- [ ] Title fallback
- [ ] Workspace avatar tooltip

Issues I noticed:

- [ ] I was not able to make the thumbnail tooltip work, but in any case, we need to translate the media type
- [ ] Pluralization is not rendering correctly for number of fact-checks:

Captura de tela de 2024-03-06 23-16-39

Feel free to just remove these two things from this PR and handle them in another PR.

src/app/components/search/SearchResultsTable/ItemThumbnail.js
```diff
@@ -32,13 +33,22 @@ const ItemThumbnail = ({
...
+      <Tooltip
+        title={
+          <FormattedMessage
+            id="ItemThumbnail.mediaType"
+            defaultMessage="{mediaType}"
+            description="type of the media item"
+            values={{ mediaType }}
+          />
+        }
+      >
```
caiosba, line 45: @danielevalverde , this still doesn't fix the problem - we actually need to map each value of `mediaType` to a different localizable string, like we do here: https://github.com/meedan/check-web/blob/develop/src/app/components/media/MediaTypeDisplayName.js. Actually, I think we can just use that component here :)

caiosba (changes requested):

caiosba (changes requested):
Sorry @danielevalverde , still seeing two issues locally:

(1) Pluralization is right, but the actual number of not displayed:

Captura de tela de 2024-03-07 13-05-56

(2) The tooltip doesn't work for social media types:

Captura de tela de 2024-03-07 13-05-38

caiosba (approved):
Yay! All good now, thanks Dani!

PR #1851, opened 2024-03-11
feed requests button

brianfleming:
## Description

Add the ability to move between regular shared feed list and INE style

References: [CV2-4376]

## Type of change

### AFTER Toggle Button Group
Screenshot 2024-03-11 at 1 50 04 PM

## Checklist

[CV2-4376]: https://meedan.atlassian.net/browse/CV2-4376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brianfleming:
ok @caiosba I think I didi this correctly, let me know if anything needs to change

caiosba:
> ok @caiosba I think I didi this correctly, let me know if anything needs to change

Thank you, I'm going to review this later today!

caiosba (changes requested):
Thanks @brianfleming, but I think there was a confusion here (correct me if I'm wrong please): since API requests can be received only when the feed is sharing published fact-checks (otherwise the API requests wouldn't return anything), the switcher should not be added to `FeedClusters` component (which is the component rendered when the feed is sharing media and requests), but to the `Feed` component, for the case when the feed is sharing fact-checks. Right? I pushed a commit that fixes it, and I also extracted the switcher as a separate component just to avoid code repetition and to make it easier for us when we hopefully delete all that code :stuck_out_tongue:

caiosba:
So, it looks good to me after these changes, feel free to merge if it looks good to you too @brianfleming :)

caiosba (approved):

brianfleming:
> Thanks @brianfleming, but I think there was a confusion here (correct me if I'm wrong please): since API requests can be received only when the feed is sharing published fact-checks (otherwise the API requests wouldn't return anything), the switcher should not be added to `FeedClusters` component (which is the component rendered when the feed is sharing media and requests), but to the `Feed` component, for the case when the feed is sharing fact-checks. Right? I pushed a commit that fixes it, and I also extracted the switcher as a separate component just to avoid code repetition and to make it easier for us when we hopefully delete all that code 😛

yeah you are right. I was hoping to have it outside of the filters, but really this should be deleted as you said, so lets just get past it. Thanks for your help clearing that up

PR #1852, opened 2024-03-12
Incorporating SimpleAnalytics to gather basic user information.

sarahkeh:
## Description

Added script tags to incorporate SimpleAnalytics.  The script tag can be included conditionally, default on (true).  Sites need to be added to SimpleAnalytics account and team members can be added from there to view and use the data.

Simple Analytics is a lightweight user tracker that doesn't rely on cookies.  Some ad blockers still block this site, so if data isn't appearing, check that the code isn't getting blocked there.  

Jira: CV2-743

## Type of change

## How has this been tested?

End to end testing, added temporary external localhost and added the site to the SimpleAnalytics app to confirm the data was being collected

Check that an ad blocker isn't blocking the analytics. config-server contains the useAnalytics tag that can turn the script tag on and off. 

## Checklist

### DIFF
--- a/src/web/views/index.js
+++ b/src/web/views/index.js
@@ -41,6 +41,13 @@ const heapAnalytics = (config) => {
     </script>` : '';
 };
 
+const simpleAnalytics = (config) => {
+  return config.useAnalytics ?
+    `<script async defer data-collect-dnt="true" src="https://scripts.simpleanalyticscdn.com/latest.js"></script>
+    <noscript><img src="https://queue.simpleanalyticscdn.com/noscript.gif" alt="" referrerpolicy="no-referrer-when-downgrade" />
+    </noscript>` : '';
+};
+
 const uptimeMonitoring = (config) => {
   return config.uptimeId ?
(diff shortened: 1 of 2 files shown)


caiosba (approved):
First Meedan PR for Sarah, thanks! :champagne:

src/web/views/index.js
```diff
@@ -41,6 +41,13 @@ const heapAnalytics = (config) => {
     </script>` : '';
 };
 
+const simpleAnalytics = (config) => {
+  return config.useAnalytics ?
+    `<script async defer data-collect-dnt="true" src="https://scripts.simpleanalyticscdn.com/latest.js"></script>
```
dariusk, line 46: The `data-collect-dnt="true"` means that if a user has "Do Not Track" enabled clientside, we will ignore it and collect data from them anyway. Is this something we want to do? CC @huslage

caiosba:
Works for me locally:

Captura de tela de 2024-03-12 16-36-02

dariusk (approved):

brianfleming:
Random question, are we going to deprecate Heap? or do we still have use for that in another way?

caiosba:
> Random question, are we going to deprecate Heap? or do we still have use for that in another way?

Both Heap and SimpleAnalytics are handled by configuration keys, so it will be up to us... but I think we should try to use only one in production.

brianfleming:
> > Random question, are we going to deprecate Heap? or do we still have use for that in another way?
> 
> Both Heap and SimpleAnalytics are handled by configuration keys, so it will be up to us... but I think we should try to use only one in production.

agreed, if we are not actively using Heap for anything, and this is the replacement, lets ensure we only have this in prod

PR #1856, opened 2024-03-15
Fix "sticky" tooltip

dariusk:
This was discovered when QA was testing CV2-4367 -- the tooltips weren't going away on buttons that entered a disabled state due to there being no more pages to go to.

This was a bug introduced in https://github.com/meedan/check-web/commit/95ca6ef12f8c3bb96197f80f0c356067c3109d95 -- we added `disableHoverListener`, which not only disables hovering, but disables the listener that tells when hovering *finishes*. So if you press "next" until the button is disabled, then mouse away from the button, the tooltip does not get removed, because the hover listener is gone. Removing this entirely, which means the tooltip will appear on the disabled button, but better than having it stick around forever.

cc @brianfleming for the design implications.


caiosba (approved):
Thanks Darius

brianfleming:
Thanks for the heads up @dariusk 
Agreed, I think the outcome of this PR is better than the bug

PR #1875, opened 2024-04-01
Applying the same filter displays different number of items

sarahkeh:
Applying the same filter displays different number of items.  To limit some confusing, it was decided that the UTC offset should be seen next to date to indicate the actual range

## Description

CV2-4416 - Applying the same filter displays different number of items

If an ISO string is converted back to a date, it assumes the local timezone unless one is stated, so using Date's getTimezoneOffset wasn't an option. The input field of the DatePicker requires an ISO value to display the correct date, so that couldn't be changed.  Plus, all saved list ranges are currently saved as ISO strings. Knowing that all start dates are made at 00:00:00 and end dates are 23:59:59, we can work backwards to figure out the original timezone the string was created in. Countries like India also have a UTC offset of +5:30 so minutes have to be considered.

## Type of change

## How has this been tested?

Checked across multiple timezones using Chrome's Sensor's feature.  

Changing timezones can change the date once the Datepicker is selected. 

## Checklist

- [X ] I have performed a self-review of my own code

### DIFF
--- a/src/app/components/search/DateRangeFilter.js
+++ b/src/app/components/search/DateRangeFilter.js
@@ -93,6 +93,7 @@ function DateRangeSelectorStartEnd(props) {
     handleChangeEndDate,
     handleChangeStartDate,
     handleClearDate,
+    getUTCOffset,
   } = props;
 
   return (
@@ -120,7 +121,7 @@ function DateRangeSelectorStartEnd(props) {
                 !valueText ?
                   <FormattedMessage id="search.anyDate" defaultMessage="any date" description="Date picker placeholder" />
                   :
-                  valueText
+                  `${valueText} ${getUTCOffset(getStartDateStringOrNull(), true)}`
               }
               onClick={onClick}
               iconRight={!valueText && <KeyboardArrowDownIcon />}
@@ -180,7 +181,7 @@ function DateRangeSelectorStartEnd(props) {
                 !valueText ?
                   <FormattedMessage id="search.anyDate" defaultMessage="any date" description="Date picker placeholder" />
                   :
-                  valueText
+                  `${valueText} ${getUTCOffset(getEndDateStringOrNull(), false)}`
               }
               onClick={onClick}
               {...params}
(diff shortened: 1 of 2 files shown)


src/app/components/search/DateRangeFilter.js
```diff
@@ -290,6 +291,30 @@ const DateRangeFilter = ({
...
+      tzSign = '+';
+      hours = 24 - hours;
+      hours -= isStart ? 0 : 1; // used for rounding up 59 minutes to 1 hr
+      hours -= isStart && minutes !== 0 ? 1 : 0; // special case for non-00 timezones
+    } else {
+      tzSign = '-';
+      hours += isStart ? 0 : 1;
+      hours += isStart && minutes !== 0 ? 1 : 0;
+    }
+    const tzOffsetMinutes = String(Math.abs(minutes)).padStart(2, '0');
```
dariusk, line 314: Wow, I didn't know about `padStart`!

dariusk (changes requested):
This looks great overall and I'm glad you're catching non-00 timezones here. I'm wondering what the input and output are actually supposed to be though. I ran this

```js
d = new Date() // Tue Apr 02 2024 10:11:20 GMT-0700 (Pacific Daylight Time)
isoDate = d.toISOString() // "2024-04-02T17:11:20.658Z" 
getUTCOffset(isoDate) // "UTC+6:48" 
```

"UTC+6:48"  is a bit weird here. Am I putting the right input string here? I am guessing from the pull request that maybe only 00:00 and 23:59 times go in, but when I run those the only result I ever get is "UTC+:0:00".

Basically, I think this points to the need for either some unit tests with potential inputs and outputs, or even more verbose comments including use cases. It's a complex function so probably unit tests would be best here.

src/app/components/search/DateRangeFilter.js
sarahkeh: Added a few unit tests using different ISO strings created in different zones using the Chrome DevTools Sensors feature.  Date will always using the current timezone of the user so fudging it in the browser is the only way to get it to spit out the correct ISO string.

amoedoamorim (approved):

PR #1883, opened 2024-04-05
Remove tags and marketplace info from shared feed create and edit

brianfleming:
## Description

As we have slowed down the Shared Feed expansion plans, we are unsure when/if we will get to the marketplace concept, as a result we will remove the Tagging and Marketplace coming soon information to simplify the interface

I have left in the output of the licenses (if applied) as it will not show up by default, and I think we will get to that eventually

References: CV2-4444

## Type of change

## How has this been tested?

manual and ran full integration test

not much

## Checklist


caiosba (changes requested):
@brianfleming , the creation is failing because actually, on the backend, "license" is mandatory:

Captura de tela de 2024-04-05 17-52-48

We can:
- Remove it from the backend as well
- Hard-coded a license in the mutation
- Set the feed as not `discoverable` in https://github.com/meedan/check-web/blob/develop/src/app/components/feed/SaveFeed.js, which disables the licenses validation on the backend

brianfleming:
> @brianfleming , the creation is failing because actually, on the backend, "license" is mandatory:
> 
> Captura de tela de 2024-04-05 17-52-48
> 
> We can:
> 
> * Remove it from the backend as well
> * Hard-coded a license in the mutation
> * Set the feed as not `discoverable` in https://github.com/meedan/check-web/blob/develop/src/app/components/feed/SaveFeed.js, which disables the licenses validation on the backend

Hmm, good catch @caiosba 
I think your option of Set the feed as not discoverable leaves us the most flexibility, what do you think?

brianfleming:
I think I did it correctly @caiosba ?? please take a look again

src/app/components/feed/SaveFeed.js
```diff
@@ -158,16 +156,11 @@ const SaveFeed = (props) => {
   const [title, setTitle] = React.useState(feed.name || '');
   const [description, setDescription] = React.useState(feed.description || '');
   const [selectedListId, setSelectedListId] = React.useState(isFeedOwner ? feed.saved_search_id : feedTeam.saved_search_id);
-  const [discoverable, setDiscoverable] = React.useState(Boolean(feed.discoverable));
+  const [discoverable] = React.useState(Boolean(false));
```
caiosba, line 159: @brianfleming , minor change, since we're not changing `discoverable`, it doesn't need to be a state - you can change this line simply to `const discoverable = false`.

src/app/components/feed/SaveFeed.js
```diff
@@ -233,25 +226,16 @@ const SaveFeed = (props) => {
...
-    if (academicLicense) licenses.push(1);
-    if (commercialLicense) licenses.push(2);
-    if (openSourceLicense) licenses.push(3);
     const input = {
       name: title,
       description,
       saved_search_id: selectedListId,
-      tags,
       licenses,
       discoverable,
```
caiosba, line 240: We can delete `discoverable` from the GraphQL query to.

caiosba (changes requested):
Thanks Brian, just a couple of minor changes - but the feature works well locally!

src/app/components/feed/SaveFeed.js
brianfleming: I guess I didn't need either, since discoverable I think is false by default. I removed both lines and the page I think works as expected

caiosba (approved):
That's even better, thanks! :)

PR #1884, opened 2024-04-06
CV2 2935 replace mui text field with cds

sarahkeh:
## Description
Convert MUI TextField to CDS TextField

Converted Files
    ProjectActions - simple conversion, no change needed
    BulkActionsAssign - simple, no change needed
    MediaActionsBar - used textarea instead of field, changed style of container to maintain width as text in textarea
 expands
    SimilarityComponent - added similarity-component-input class to made input inline with text, kept small variable
    ThresholdControl - added similarity-component-input to keep input inline and spaced with Slider
    SmoochBotIntegrationButton - simple, no change needed
    SmoochBotIntegrations - simple, no change needed
    SmoochBotMenuOption - simple, no change needed
    SmoochBotMenuOption - simple, no change needed
    SmoochBotTextEditor - simple, no change needed
    TranslateStatuses - simple, no change needed

Unconverted: 
     RuleField - Uses MUI Autocomplete, doesn't work with our TextField

References: CV2-2935

## Type of change

## How has this been tested?

Inputs compared to original to verify similar appearance, ensured input type and limitations are the same.  

Lots of changes, most on Similarity 

## Checklist


caiosba:
Thanks Sarah! I suggest you also run integration tests against your branch, as explained here: https://meedan.atlassian.net/browse/CV2-4396. TL;DR: Push a commit (it can be a blank commit) with `[full ci]` in the commit message (for example: `git commit --allow-empty -m '[full ci] Run all integration tests for this branch'`).

brianfleming (commented):
Just a minor comment @sarahkeh that it could be good to cleanup unused attributes when making these changes as well. For example In `TranslateStatuses.js` you can remove the `fullWidth` attribute from the TextField since the CDS version doesn't use it. There are others, just using that one as an example

src/app/components/cds/inputs/TextField.js
```diff
@@ -36,6 +36,7 @@ const TextField = React.forwardRef(({
   placeholder,
   required,
   variant,
+  size,
```
brianfleming, line 39: @sarahkeh can you give some context on adding this size? Since our design system doesn't have different sizes for TextField I hesitate to add this in the codebase. If its to solve an issue on a particular screen, maybe there is a layout change to be made on that screen instead of adding this.

sarahkeh: It's used on Similarity and translated Statuses page only, just shrinks the height of the input field a little.  I could easily make the change on the screen instead, that way it wouldn't touch the main component.  (also, sorry, didn't notice your second comment until now)

brianfleming: @sarahkeh thanks!
If you mean that a place like this (statuses in settings)
Screenshot 2024-04-08 at 8 45 37 PM

That the input is slightly smaller. I would just ignore it completely. I would rather have the CDS textfield input be consistent with the design system and not worry about adding this size attribute. We don't need the maintain the old layout, using the design system more should help everything look consistent.

If when you take size out, if you have a concern about a particular layout, just let me know and we can come up with another solution for those places.

sarahkeh: Alright, I think the default TextField looks fine, should be good with the base styling, but if you feel differently I can update it.  
Screenshot 2024-04-15 at 10 38 46 AM

brianfleming: Thanks @sarahkeh looks good to me. I would just run integration tests again to make sure you got them all fixed up

PR #1889, opened 2024-04-10
Updates from translation feedback

brianfleming:
## Description

Update to translation strings based on feedback on transifex

## Type of change

## How has this been tested?

manual testing

not much, this should be mainly text updates

## Checklist

### DIFF
--- a/localization/react-intl/src/app/components/cds/media-cards/MediaCount.json
+++ b/localization/react-intl/src/app/components/cds/media-cards/MediaCount.json
@@ -1,7 +1,7 @@
 [
   {
     "id": "sharedItemCard.medias",
-    "description": "A count of medias in an item. Title-case where applicable. Example: 3 Medias",
-    "defaultMessage": "{mediaCount, plural, one {# Media} other {{mediaCount, number} Medias}}"
+    "description": "A count of media in an item.",
+    "defaultMessage": "{mediaCount, number} Media"
   }
 ]
--- a/localization/react-intl/src/app/components/cds/media-cards/MediaCount.json
+++ b/localization/react-intl/src/app/components/cds/media-cards/MediaCount.json
@@ -1,7 +1,7 @@
 [
   {
     "id": "sharedItemCard.medias",
-    "description": "A count of media in an item.",
-    "defaultMessage": "{mediaCount, number} Media"
+    "description": "A count of media in an item. Title-case where applicable. Example: 3 Media. In English we prefer Media for both singular and plural",
+    "defaultMessage": "{mediaCount, plural, one {# Media} other {{mediaCount, number} Media}}"
   }
 ]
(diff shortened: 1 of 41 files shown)


src/app/components/search/SearchResultsTable/ItemThumbnail.js
```diff
@@ -28,6 +28,7 @@ const ItemThumbnail = ({
     if (picture) {
       return (
         <Tooltip
+          arrow
```
brianfleming, line 31: reviewers FYI, just threw this in there because I noticed it missing

localization/react-intl/src/app/components/cds/media-cards/MediaCount.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "id": "sharedItemCard.medias",
-    "description": "A count of medias in an item. Title-case where applicable. Example: 3 Medias",
-    "defaultMessage": "{mediaCount, plural, one {# Media} other {{mediaCount, number} Medias}}"
```
amoedoamorim, line 5: I'm not sure this is a good change as it's probably worth having clear singular/plurals in other languages.

brianfleming: @amoedoamorim I wasn't actually sure how to handle this, maybe you can help.
I changed this, because in English Media/Medias as both "correct" plurals, but to most native english speakers Medias sounds wrong to the ear. So I would prefer in english to use Media for both singular or plural, but wasn't sure how to do this for other languages.

Should I just revert this change and put Media for both situations? and then the other languages can be translated for themselves?

amoedoamorim: > Should I just revert this change and put Media for both situations? and then the other languages can be translated for themselves?

Yeah, sounds like the best approach here

localization/react-intl/src/app/components/media/BulkActionsRemoveTag.json
```diff
@@ -32,7 +32,7 @@
   {
     "id": "bulkActionsRemoveTag.dialogBody",
     "description": "Body of dialog warning that bulk untagging items",
-    "defaultMessage": "The <strong>{tags}</strong> tag will be removed from selected items that have this tag. if a tag is not found in a selected item, the item will be unaffected.<br /><br /><strong>This action cannot be undone.</strong> Are you sure you want to continue?"
+    "defaultMessage": "The <strong>{tags}</strong> tag will be removed from selected items.<br /><br /><strong>This action cannot be undone.</strong> Are you sure you want to continue?"
```
amoedoamorim, line 35: I noticed the {tags} variable can be a single tag or a list of tags. Perhaps it can read "The {tags} tags will be removed..." and even if the variable gets replaced by a single tag name it still makes sense as it refers to all instances of it on selected items?

brianfleming: good idea, done....who are you? the new AO? 😁

amoedoamorim (approved):

PR #1912, opened 2024-04-26
Show imported item requests

caiosba:
## Description

When media items are imported from a shared feed, they become available in the target workspace. When the user clicks on those imported media, a dialog opens with a toggle, which the user can use to switch between requests imported with those media or requests that the media received through their tipline.

References: CV2-4490 (feature) and CV2-4459 (bug fix).

## Type of change

## How has this been tested?

See related ticket CV2-4490 for detailed test plan.

I had to deal with very old and legacy Relay Classic components, so I did my best to touch those as least as possible, since they are part of crucial features. I didn't bother to refactor or fix any legacy code and tried to concentrate most of my work in new components, and that was intended, since I wanted to minimize the risk of breaking things.

## Checklist


caiosba:
@amoedoamorim please prioritize reviewing this.

src/app/components/feed/FeedItemMediaDialog.js
```diff
@@ -34,7 +34,7 @@ const FeedItemMediaDialog = ({ itemDbid, classes }) => {
         variables={{
           currentTeamSlug,
           feedDbid: parseInt(feedDbid, 10),
-          projectMediaDbid: parseInt(projectMediaDbid, 10),
+          projectMediaDbid: parseInt(itemDbid, 10),
           itemDbid: parseInt(itemDbid, 10),
```
amoedoamorim, line 38: Why are there 2 variables with same value?

caiosba: Yeah, let me combine the two. The explanation is:

- A cluster contains multiple items, let's say, A, B and C
- You want to get the cluster which item B belongs to
- From that cluster, you can get the item C

Since clusters are ephemeral, the API doesn't allow clusters to be retrieved by their ID. Previously this component was being used for different IDs, but now they are always the same, for simplicity.

amoedoamorim (approved):
LGTM

PR #1918, opened 2024-05-01
Tweaks to WorkspaceItemCard and Trash rendering

dariusk:
This fixes a few issues found with the trash view:

 - text alignment on the item date
 - add back the "Select All" checkbox
   - update `SelectAllTh` to accept class overrides
 - add last request date, suggestions count, and the publication checkbox
 - display the localized text for an item's status rather than the unlocalized label

The major remaining issue here is that the trash view doesn't pass a "channels" value to `WorkspaceItemCard`. This is the array of objects specified in the `propTypes` and you can see an example of it in action in the `FeedClusters` component. I believe this will require some kind of graphql subquery we don't have in place right now, but getting it to render once we have the data should be identical to how it's done in `FeedClusters`, using `CheckChannels` to map from ID to string for each item in the array.

References: CV2-3991

## Type of change

## Checklist


brianfleming:
@amoedoamorim I think we should merge this before we revert the search results part, just to get these fixes in to the cards

amoedoamorim (approved):
Will merge now and then revert relevant  `SearchResults.js` changes to restore table view in Trash

PR #1923, opened 2024-05-03
 CV2-4477 new Slideout component

sarahkeh:
## Description

Slideout component requires a title (object or string) and an onClose function. Footer has a boolean flag and all its contents are optional.
Contents can be Close, MainActionButton, SecondaryActionButton, and OptionalNode. Content can contain anything and is scrollable.
Slideout works in rtl and ltr layouts.

CV2-4477
## Type of change

## How has this been tested?

Unit tested and manually tested in the Sandbox

Noticed with 3 or more buttons, the footer tends to become squished in smaller screens, thus the choice to use small buttons in the Sandbox.  Content was tested with input fields and large blocks of text.  

## Checklist

### DIFF
--- a/src/app/components/cds/slideout/Slideout.js
+++ b/src/app/components/cds/slideout/Slideout.js
@@ -13,6 +13,7 @@ const Slideout = ({
   footer,
   onClose,
   showCancel,
+  cancelProps,
   mainActionButton,
   secondaryActionButton,
   optionalNode,
@@ -36,9 +37,9 @@ const Slideout = ({
 
   return (
     <div className={cx(styles.slideoutMain, 'slideout', isOpen ? styles.slide : null)}>
-      <div className={cx(styles.slideoutContainer)}>
-        <div className={cx(styles.slideoutTitle)}>
-          <div className={cx(styles.slideoutTitleContent)}>
+      <div className={styles.slideoutContainer}>
+        <div className={styles.slideoutTitle}>
+          <div className={styles.slideoutTitleContent}>
             <h6>
               {title}
             </h6>
@@ -50,13 +51,13 @@ const Slideout = ({
             />
           </div>
         </div>
-        <div className={cx(styles.slideoutBody, footer ? styles.withFooter : null)}>
+        <div className={styles.slideoutBody}>
           {content}
         </div>
         { footer &&
-          <div className={cx(styles.slideoutFooter)}>
-            <div className={cx(styles.slideoutFooterMain)}>
-              <div className={cx(styles.slideoutFooterLeftContent)}>
+          <div className={styles.slideoutFooter}>
+            <div className={styles.slideoutFooterMain}>
+              <div className={styles.slideoutFooterContent}>
                 { showCancel &&
                   <ButtonMain
                     variant="text"
@@ -64,11 +65,12 @@ const Slideout = ({
                     onClick={() => handleClose()}
                     label={
                       <FormattedMessage
-                        id="label.SlideoutCancelButton"
-                        description="Label of slideout cancel button"
+                        id="global.cancel"
                         defaultMessage="Cancel"
+                        description="Regular Cancel action label"
                       />
                     }
+                    {...cancelProps}
                   />
                 }
                 { mainActionButton &&
@@ -83,7 +85,7 @@ const Slideout = ({
                 }
               </div>
               { secondaryActionButton &&
-                <div className={cx(styles.slideoutFooterRightContent)}>
+                <div className={styles.slideoutFooterContent}>
                   {secondaryActionButton}
                 </div>
               }
@@ -100,6 +102,7 @@ Slideout.defaultProps = {
   content: null,
   footer: true,
   showCancel: false,
+  cancelProps: {},
   mainActionButton: null,
   secondaryActionButton: null,
   optionalNode: null,
@@ -111,6 +114,7 @@ Slideout.propTypes = {
   footer: PropTypes.bool,
   onClose: PropTypes.func.isRequired,
   showCancel: PropTypes.bool,
+  cancelProps: PropTypes.object,
   mainActionButton: PropTypes.instanceOf(ButtonMain),
   secondaryActionButton: PropTypes.instanceOf(ButtonMain),
   optionalNode: PropTypes.node,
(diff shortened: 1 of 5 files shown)


src/app/components/cds/slideout/Slideout.module.css
```diff
@@ -0,0 +1,94 @@
...
+    display: flex;
+  }
+
+  .slideoutFooterRightContent {
+    align-items: flex-end;
+    display: flex;
+  }
+}
+
+body[dir='rtl'] {
```
brianfleming, line 74: its more performant css to animate something like `transform: translate3d(-600px, 0, 0);` instead of `right` and `left`

src/app/components/cds/slideout/Slideout.js
```diff
@@ -0,0 +1,119 @@
...
+            <div className={cx(styles.slideoutFooterMain)}>
+              <div className={cx(styles.slideoutFooterLeftContent)}>
+                { showCancel &&
+                  <ButtonMain
+                    variant="text"
+                    theme="text"
+                    onClick={() => handleClose()}
+                    label={
+                      <FormattedMessage
+                        id="label.SlideoutCancelButton"
```
brianfleming, line 67: I think this is backwards, or not following our normal pattern? @amoedoamorim correct?

brianfleming (changes requested):
looking good @sarahkeh ! 🔆
I put a couple small comments in the code, nothing major,
but one layout thing to clean up is the footer buttons/nodes/etc
Check out these two screenshots:

Screenshot 2024-05-03 at 3 54 18 PM
Screenshot 2024-05-03 at 3 54 48 PM

Since the "Cancel" button is IN the slideout component, but the extra buttons and nodes are not, the cancel button is default size and the other passed in are small size, its showing that they are not vertically aligned.

I think its a good implementation that the extra nodes can be self contained, but as a result, can you update the footer css to ensure that things are vertically aligned?

Also, I would give the footer containers a gap value of like 8px, so that way each element gets some space around

brianfleming:
I noticed one other things @sarahkeh that I didn't account for in the designs, but should be a simple change. and really its in our control, but because of translations this may happen in some languages, BUT, if you put in a very long slideout title, you get:

Screenshot 2024-05-03 at 4 11 06 PM

Can you update the css to truncate and ... the title in these cases, so it results in something like:
Screenshot 2024-05-03 at 4 13 19 PM

src/app/components/cds/slideout/Slideout.js
amoedoamorim: yeah, @sarahkeh the usual naming pattern for string ids is: "componentName.whatIsItAbout"

For a cancel button, though, we don't need to translate this string yet again. Please reuse the values in `id`,  `description` and `defaultMessage` from cancel in `globalStrings.js`. Just copy and paste the same values to make it work, please.

src/app/components/cds/slideout/Slideout.js
```diff
@@ -0,0 +1,119 @@
...
+  const handleClose = () => {
+    setIsOpen(false);
+    setTimeout(() => {
+      onClose();
+    }, 390);
+  };
+
+  return (
+    <div className={cx(styles.slideoutMain, 'slideout', isOpen ? styles.slide : null)}>
+      <div className={cx(styles.slideoutContainer)}>
```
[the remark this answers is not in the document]
sarahkeh: That's good to know! I've removed the unnecessary uses of cx so it's a bit cleaner.

src/app/components/cds/slideout/Slideout.module.css
sarahkeh: I didn't know about this, thanks for telling me! Been using framer motion for too long, forgot about the basics. Updated

src/app/components/cds/slideout/Slideout.js
sarahkeh: Replaced this with global cancel message

sarahkeh:
@brianfleming All the updates should now be in the latest commit, let me know if you see anything else!

src/app/components/cds/slideout/Slideout.js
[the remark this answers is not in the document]
brianfleming: Awesome. classnames is a pretty powerful package for applying multiple classes and conditional classes etc. Their docs are pretty good, and if you look at our ButtonMain component that has a fair bit of use of it which could be helpful to look at.

src/app/components/cds/slideout/Slideout.module.css
```diff
@@ -0,0 +1,87 @@
...
+  position: fixed;
+  top: 0;
+  width: 25vw;
+}
+
+.slideoutContainer {
+  background-color: var(--otherWhite);
+  display: flex;
+  flex-direction: column;
+  max-height: 100%;
```
brianfleming, line 15: update this to height instead of max-height

src/app/components/cds/slideout/Slideout.module.css
```diff
@@ -0,0 +1,87 @@
...
+    h6 {
+      margin: 0;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+  }
+}
+
+.slideoutBody {
```
brianfleming, line 40: you could add `flex: 1 1 auto;` to get this to expand to fill the remaining space

brianfleming (changes requested):
Thanks for those changes @sarahkeh I added one unrelated css change in a comment, but another thing to address is if the content of the slideout is very small, the slideout body is not expanding to fill the space. See the screenshot below, but the footer should always be at the bottom:

Screenshot 2024-05-07 at 1 01 05 PM

I will put in some css comments as a way to address this

src/app/components/cds/slideout/Slideout.module.css
sarahkeh: Updated, content box no longer shrinks with small amounts of content

brianfleming (approved):
Looking great @sarahkeh Thanks for the updates.
Two small notes:
- probably worth a real engineers approval review on this in addition to just me.
- in the future don't forget to re-request a review so I get notified to look at it again when you are ready with changes

slide

amoedoamorim (approved):

PR #1941, opened 2024-05-10
Deprecate styled components

brianfleming:
## Description

We only had a few references left to styled-components. This PR removes the last few, transitions them to css files and removes the styled-component packages.

## Type of change

## How has this been tested?

Updated all the necessary test and manually verified all the changes visually

Not much, things should look exactly the same

## Checklist


src/app/components/media/MediaCardLarge.js
```diff
@@ -55,85 +46,86 @@ const MediaCardLarge = ({
...
-          </Box>
-        ) : null }
-        { type === 'UploadedImage' ? (
-          <ImageMediaCard
-            projectMedia={projectMedia}
-            imagePath={media.embed_path}
-            currentUserRole={currentUserRole}
-            superAdminMask={superAdminMask}
+    <div
+      inModal={inModal}
```
amoedoamorim, line 50: inModal isn't helpful here anymore

brianfleming: 🙃 uhhh..yeah true. removed!

src/app/components/source/SourcePicture.js
```diff
@@ -154,13 +115,22 @@ class SourcePicture extends Component {
...
+      <div
         alt="avatar"
         size={this.props.size}
         type={this.props.type}
-        className={this.props.className}
+        className={cx(
+          styles['source-picture'],
+          {
+            [this.props.className]: true,
+            [styles.sizeLarge]: this.props.size === 'large',
```
amoedoamorim, line 126: This is more of a question than a review comment, but how do you approach style class naming? I'm not always sure when to name my class using camelCase, PascalCase or kebab-case

brianfleming: yeah, honestly its something we/I am not consistent about, but, but maybe as a front-end team we can come to some sort of standard.
Random not complete thoughts that don't lead to any conclusion, but 🤷 :
- sometimes when doing refactoring I maintain "whatever was there before" to keep the PR clearer
- camelCase is more like our js, so sometimes it more closely matches the other parts of the code
- kebab-case is more of a css standard approach that most people prefer/linters

amoedoamorim (approved):

PR #1982, opened 2024-06-22
Epic/3939 run updated selenium image on travis

amoedoamorim:
## Description

This PR updates our selenium standalone chrome image and does a few tweaks to ensure stability.

It reduces the selenium server-side timeout (120s) from the image default (300s), increases the client-side timeout for new sessions (180s) and open more concurrent session slots so that tests don't fail sequentially with `Net::ReadTimeout` due to stuck not-cleaned up slots.

Due to the new chrome cookie handling it also requires that test scripts running on web container and the application running on the browser see everything served from the same domain, so we're adding nginx to the web container to make it all appear to be running on localhost from whichever container.

It also fixes the imgur upload, which was failing with "400: Bad Request". Changed the type to `base64` and it seems to work now.

H/T @caiosba @danielevalverde for the help taming this selenium monster

References: CV2-3939

## Type of change

## How has this been tested?

Tested the setup locally, manually starting the nginx proxies on web and chromedriver before running local integration tests.

- Etc.

## Checklist

- [x] I have made needed changes to the README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,6 +85,8 @@ jobs:
       name: similarity-tests
       if: branch = develop OR branch = master OR commit_message =~ /\[full ci\]/ OR commit_message =~ /\[similarity tests\]/
       script:
+      - docker-compose exec web service nginx start
+      - docker-compose -f docker-compose.yml -f docker-test.yml exec chromedriver service nginx start
       - docker-compose exec
         -e IMGUR_CLIENT_ID=$IMGUR_CLIENT_ID
         -e TRAVIS_JOB_NAME=$TRAVIS_JOB_NAME
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,10 +70,7 @@ jobs:
       script:
       - docker ps
       - docker-compose exec web service nginx start
-      - docker-compose exec web curl http://localhost:3000 # test web nginx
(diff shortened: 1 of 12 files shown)


caiosba:
I'm going to review it tomorrow, but for now I would like to send big kudos for the amazing work handling this beast, @amoedoamorim ! 👏🏽👏🏽👏🏽

.travis.yml
```diff
@@ -51,8 +51,8 @@ before_install:
 - cp ../test/google_credentials.json google_credentials.json
 - cd -
 # Configuration files
-- cp test/config.js.example config.js
-- cp test/config.js.example test/config.js
+- cp config.js.example config.js
+- cp config.js.example test/config.js
```
caiosba, line 55: Does it mean that we don't need the `example` files inside the `test/` directory? If so, can we delete them?

caiosba (approved):
Great work @amoedoamorim! :tada:

danielevalverde (approved):

.travis.yml
amoedoamorim: Hmm, we still do need `test/config.yml.example` and we might still use `test/config.js.example` if we're not running the nginx proxies locally, which I haven't made autostart for the local test env. It gets started automatically for Travis jobs.

caiosba: OK sounds good!

PR #1985, opened 2024-06-25
Cv2 4498 edit article

sarahkeh:
## Description

Adds edit to the create form

Info added to top of form, Saves on blur, added cds component for Rating dropdown to display fact check rating, updated ArticleCard to take in as props additional information that the edit form will need.  This might need to be added to as these components are properly hooked up 

References: CV2-4498

## Type of change

## How has this been tested?

Added a rough onClick even to open up the editing form, this should be properly connected as now it also opens the form when the user tries adding a tag.  

## Checklist


brianfleming (changes requested):
Hi @sarahkeh I starting looking at this and have some things to address. I will continue to take a look, but didn't want to hold you up looking into somethings:

- [x] The created and edited areas are not showing the user information of who created or saved, I just see the dates:
Screenshot 2024-06-25 at 2 24 03 PM

- [x] As discussed on a call at some point, we are going to skip trash for now, so that should be removed from the footer
- [x] There is no validation happening on required fields right now. The components should all have an ability to get the error state when appropriate
- [x] Just like you did with the "new" button interaction, we need to close and open each article when clicked from the list, currently the slideouts will "stack", so we should match the behavior

I will keep looking, but as said, wanted to get somethings in front of you. Let me know if you have any questions or if I can clarify anything!

danielevalverde (changes requested):
Hi, @sarahkeh  :) 

I noticed that when clicking on the "Unpublished Report" button from the forms, it redirects to the wrong URL:  
Captura de tela_2024-06-25_17-51-31

Could you please update the handleGoToReport function to redirect to the correct URL from the list page as well?

src/app/components/article/ArticleForm.js
```diff
@@ -22,20 +25,19 @@ const ArticleForm = ({
...
+      {articleType === 'fact-check' && mode === 'edit' && <FormattedMessage id="article-form-fact-check-edit-title" defaultMessage="Edit Claim & Fact-Check" description="Title for the slideout edit fact check form" />}
     </>
   );
 
-  const [claimDescription, setClaimDescription] = React.useState(articleType === 'fact check' && projectMedia?.suggested_main_item ? projectMedia.suggested_main_item.claim_description : '');
-  const [claimContext, setClaimContext] = React.useState('');
-  const options = team.team?.tag_texts?.edges.map(edge => ({ label: edge.node.text, value: edge.node.text })) || [];
+  const [claimDescription, setClaimDescription] = React.useState(article?.claim_description?.description || '');
+  const [claimContext, setClaimContext] = React.useState(article?.claim_description?.context || '');
+  const options = team.team?.tag_texts?.edges.map(edge => ({ label: edge.node.text, value: edge.node.text })) || team.teamTags.map(tag => ({ label: tag, value: tag }));
```
amoedoamorim, line 40: question: is this supposed to be `team.team`?

src/app/components/search/SearchResultsCards/ArticleCard.js
```diff
@@ -26,56 +31,95 @@ const ArticleCard = ({
...
+          rating,
+          createdDate,
+          statuses,
+          id,
+          claim_description,
+          publishedAt,
+        }}
+      />}
+    </div>
+  );
```
amoedoamorim, line 120: Issue (blocking): I think we should decouple this from the `ArticleCard` component and maybe have these form component mounted elsewhere, probably the `Articles` component, that should manage the `openEdit` state and so on. We should pass a callback from the parent `Articles` to `ArticleCard` `onClick` prop that will trigger the open form state.

amoedoamorim (changes requested):
Please mount the forms from the Articles component rather than from inside each card

src/app/components/article/Articles.js
```diff
@@ -89,6 +94,14 @@ const ArticlesComponent = ({
     });
   };
 
+  const handleClick = (article) => {
+    console.log('Click!'); //eslint-disable-line
```
brianfleming, line 98: I appreciate the enthusiasm of this, but it probably should be removed

danielevalverde (changes requested):
Hi, Sarah :) 

i am getting this error when clicking on an article on list view:
`ArticleForm.js:40 Uncaught TypeError: Cannot read properties of null (reading 'map') at ArticleForm (ArticleForm.js:40:1)`
Captura de tela_2024-07-04_18-09-14
I believe It is happening because my team doesn't have tags, it can be fixed by adding a safe navigation on this part of the code:
`team.teamTags?.map` https://github.com/meedan/check-web/blob/cv2-4498-edit-article/src/app/components/article/ArticleForm.js#L40 

And I'm not sure how to test the edit forms. I tried editing an article and closing the form, but the changes are not being saved.

brianfleming (commented):
Couple things I wanted to check on @sarahkeh and let the others handle technical review:

- There are a lot of relay warnings that get generated when the edit slideout opens, i just wanted to flag these to make sure they are not important:
such as: `Warning: RelaySelector: Expected object to contain data for fragment `ArticleForm_team`, got `...`
- Editing the Status of the Fact-Check doesn't seem to work, it never updates. Is this a bug or a know thing?

brianfleming (changes requested):
@sarahkeh something to address other than my previous questions, we should remain consistent with the slideouts and click interactions. So a couple things I noticed, and you can see them in the attached video:
- Currently when editing, I can only open 1 slideout and any other clicks do nothing. We should have it behave like the create button does, where it replaces the previous.
- Conversely, If you open an article to edit, then click the create button, it stacks instead of replacing:


https://github.com/meedan/check-web/assets/418001/ece36fdf-b67b-4f35-9325-b23d61e5a82b

sarahkeh:
> * Editing the Status of the Fact-Check doesn't seem to work, it never updates. Is this a bug or a know thing?

@brianfleming I think what's happening with the Status is that it went from a value tied to a media item to a value called Rating that is tied to a Fact Check.  What the form does is update and save the Fact Check Rating, but I think the list displays the media item's Status, not the Rating.  I thought @caiosba said these values were synced but maybe that work is pending?

caiosba:
Yes, the form should set and read the `FactCheck` `rating`. If the media one displayed in the card or item page then this is something on the backend side that we can check and fix. Please file separately if that's the case.

sarahkeh:
> @sarahkeh something to address other than my previous questions, we should remain consistent with the slideouts and click interactions. So a couple things I noticed, and you can see them in the attached video:
> 
> * Currently when editing, I can only open 1 slideout and any other clicks do nothing. We should have it behave like the create button does, where it replaces the previous.
> * Conversely, If you open an article to edit, then click the create button, it stacks instead of replacing:

@brianfleming I've just pushed the change for the top point but the second is much more complicated.  I can prevent the create form opening over the edit and vice versa (by checking for the class name) but I can't close a Create form from inside the Articles List as that's a value that exists within the Create New Article component.

sarahkeh:
> Yes, the form should set and read the `FactCheck` `rating`. If the media one displayed in the card or item page then this is something on the backend side that we can check and fix. Please file separately if that's the case.

@caiosba I created CV2-4901 to work on this issue

brianfleming (commented):
Ok, I will let the engineers approve or ask for changes. the reality is that this is getting very long lived and we need to get moving. With the creation of the rating ticket to be addressed which is currently the most confusing part for users, lets get this into the epic branch so we can easily evaluate the completness of this feature set

caiosba:
Agree with @brianfleming and aligned with @amoedoamorim that he will do the tech review soon.

sarahkeh:
Changes that have been added:

- Fixed link for publishing a report
- Added safe nav for tags
- Removed form from ArticleCard component, added to Articles componenet
- Clicking another article will swap the slideout rather than having the user have to close the current edit slideout first
- team.team removed 
- added user name to Last Saved, removed created by because no user is provided. 
- Trash removed for now, footer also hidden
- Required field validation

src/app/components/article/ArticleForm.js
sarahkeh: was passing a prop incorrectly, now it's just `team`

amoedoamorim (approved):
:lgtm:

PR #2005, opened 2024-07-17
Moar articles review cleanup

brianfleming:
## Description

- Cleanup empty state on tags settings list
- Simplify the privacy requests area, no need for buttons
- Add tooltip on remove articles button in list on media cluster page
- Move the count of selected items in a list for bulk changes to the bulk actions button for clarity
- Add cancel button to existing article slideout
- Make footer default false for slideout with no actions in footer
- Add an indicator like the side navigation to the articles list to show when the area is scrolling
- Stop hiding side navigation on media item cluster pages now that we have 2 columns

References: CV2-4441 , CV2-4874

## Type of change

## How has this been tested?

reviewing

### TAGS EMPTY BEFORE
Screenshot 2024-07-17 at 9 30 32 AM

### TAGS EMPTY AFTER
Screenshot 2024-07-17 at 9 34 51 AM

### PRIVACY AFTER
Screenshot 2024-07-17 at 10 05 35 AM

## Checklist

### DIFF
--- a/src/app/components/cds/slideout/Slideout.js
+++ b/src/app/components/cds/slideout/Slideout.js
@@ -111,7 +111,7 @@ const Slideout = ({
 
 Slideout.defaultProps = {
   content: null,
-  footer: true,
+  footer: false,
   showCancel: false,
   cancelProps: {},
   mainActionButton: null,
--- a/localization/react-intl/src/app/components/article/ArticleForm.json
+++ b/localization/react-intl/src/app/components/article/ArticleForm.json
@@ -67,7 +67,7 @@
   {
     "id": "articleForm.noClaimDescript",
     "description": "Message overlay to tell the user they must complete additional fields to unlock this area",
-    "defaultMessage": "Start by adding Claim information for this Fact-Check"
+    "defaultMessage": "Start by adding Claim information<br />for this Fact-Check"
   },
   {
     "id": "mediaFactCheck.factCheck",
--- a/localization/react-intl/src/app/components/article/RemoveArticleButton.json
+++ b/localization/react-intl/src/app/components/article/RemoveArticleButton.json
@@ -10,6 +10,11 @@
     "defaultMessage": "Article successfully removed from item."
   },
   {
+    "id": "removeArticleButton.tooltip",
+    "description": "Tooltip message displayed on remove article button.",
+    "defaultMessage": "Remove article from media cluster"
+  },
+  {
     "id": "removeArticleButton.confirmationDialogTitle",
     "description": "Title displayed on a confirmation modal when a user tries to remove an article.",
     "defaultMessage": "Are you sure you want to remove this article?"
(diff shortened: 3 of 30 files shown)


brianfleming:
I need a little help with one last thing on this PR from reviewers. Not sure what is going on, let me describe. I changed the behavior to stop hiding the side navigation on media item cluster pages because now that we have 2 columns we have enough room for everything and we planned on doing this in the future anyway. However, it causes a crash in one case:
STR:
- Click on an item in a tipline list
- use the top forward nav arrow at the top of the item page to go to the next item in the list
- click on "All Items" in the now visible side navigation
- Result: crash

Note: if you do that ^^ and click back to any other list, like inbox or a custom list, the crash does not occur. it seems like its only with "All Items"

src/app/components/cds/slideout/Slideout.js
```diff
@@ -111,7 +111,7 @@ const Slideout = ({
 
 Slideout.defaultProps = {
   content: null,
-  footer: true,
+  footer: false,
```
brianfleming, line 114: @sarahkeh somethign to get your eyes on specifically is this change to the slideout that I made. I noticed that if there is a slideout with no buttons or info, then the footer is still shown, just empty. So i figured its better to be opt-in instead of out.

amoedoamorim: Also, does `footer` need to be its own boolean prop or instead showing the footer or not can/should be simply inferred from the existence of footer content?

brianfleming: probably better answered/addressed by @sarahkeh as the author of that component, but seems reasonable

amoedoamorim (approved):
:lgtm:

brianfleming:
Thanks @amoedoamorim I will more forward after @sarahkeh weighs in on the slideout footer thing

src/app/components/cds/slideout/Slideout.js
sarahkeh: It could probably be removed, my thinking was it could be easier in some instances where the slideout contains several buttons but might need to be hidden for certain users or in certain states to just set one value. But I'm not sure that's necessary, could definitely update it to just check if any of the buttons are there.

amoedoamorim: Sounds good. That's more of a suggestion than a requirement. If it's going to be too much trouble, let it be

sarahkeh:
> Ooops, I forgot about my own comment about the All Items crash thing before merging. @amoedoamorim did you get a chance to look at that?

It looks like it's happening when the ProjectMedia query and the Search query get batched together.  When that query returns is when the crash happens. Maybe it's trying to set ProjectMedia values for a component that no longer exists?

caiosba:
Hum... I'm not seeing this crash locally... any ideas of how to reproduce it?

caiosba:
> Ooops, I forgot about my own comment about the All Items crash thing before merging. @amoedoamorim did you get a chance to look at that?

@amoedoamorim FYI we handled it in conversations.

PR #2069, opened 2024-08-23
localization changes because of linter

brianfleming:
@sarahkeh and @amoedoamorim I think all these localization changes that get flagged as changed when you restart the app on develop is because of the eslint PR. If I remember correctly you took them out in order to review easier?

If thats so, I figured it would be best to have this standalone PR with just those changes. amirite?

### DIFF
--- a/localization/react-intl/src/app/components/article/ArticleForm.json
+++ b/localization/react-intl/src/app/components/article/ArticleForm.json
@@ -80,9 +80,9 @@
     "defaultMessage": "Fact-check"
   },
   {
-    "id": "articleForm.reportPublishedTitle",
-    "description": "Title of alert box in article form.",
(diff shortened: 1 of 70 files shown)


sarahkeh:
Agreed, although I'm not sure if I should be the one to approve this PR given that it's like my linter work part 2.

brianfleming:
> Agreed, although I'm not sure if I should be the one to approve this PR given that it's like my linter work part 2.

sounds good. We can let @amoedoamorim make the call on this one

amoedoamorim:
Thanks Brian and Sarah. Reviewing this now. I'll also push another commit which will be the compiled Webstructured.json resource.

amoedoamorim (approved):

PR #2076, opened 2024-08-27
add back missing done label to multiselect

brianfleming:
Thanks for catching this @amoedoamorim here is the label back


sarahkeh (approved):
👍

PR #2031, opened 2024-08-02
Refactoring the "Manage Media" button on the item page.

caiosba:
## Description

Replace the "Manage Media" drop-down by a single "Merge Items" buttons, with copy changes.

Reference: CV2-5014.

## Type of change

- [x] Performance improvement and/or refactoring (non-breaking change that keeps existing functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Security mitigation or enhancement
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Automated test (add or update automated tests)

## How has this been tested?

Manually.

## Things to pay attention to during code review

CV2-5014-2-ezgif com-video-to-gif-converter

## Checklist

- [x] I have performed a self-review of my own code
- [x] I've made sure my branch is runnable and given good testing steps in the PR description
- [x] I considered secure coding practices when writing this code. Any security concerns are noted above.
- [ ] I have commented my code in hard-to-understand areas, if any
- [ ] I have made needed changes to the README
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] If I implemented any new components, they are self-contained, their `propTypes` are declared and they use React Hooks and, if data-fetching is required, they use Relay Modern with fragment containers
- [ ] If my components involve user interaction - specifically button, text fields, or other inputs - I have added a [BEM-like class name](https://meedan.atlassian.net/wiki/spaces/ENG/pages/1327628289/Naming+conventions+for+interactive+elements) to the element that is interacted with
- [x] To the best of my knowledge, any new styles are applied according to the design system
- [ ] If I added a new external dependency, I included a rationale for doing so and an estimate of the change in bundle size (e.g., checked in https://bundlephobia.com/)

PR #2078, opened 2024-08-28
Merge Media Choice

brianfleming:
## Description

Re-introduce the concept of importing and exporting media between clusters. The difference in this implementation is some wording and where the choice happens. It now happens in the dialog.

The default is export in order to match the Factly workflow, which is more aligned with how most partners probably behave. This also helps with processing the inbox

References: CV2-5156

## Type of change

## How has this been tested?

For reference on the changes this is the PR that we are "sort of reverting"
https://github.com/meedan/check-web/pull/2031

### Toggle Button Choice for merge actions
Screenshot 2024-08-30 at 9 56 23 AM

### Disabled with tooltip
Screenshot 2024-08-30 at 9 57 36 AM

## Checklist


brianfleming:
@caiosba when you take a look, i copied a lot of the naming conventions from the old code in [the previous pr](https://github.com/meedan/check-web/pull/2031).

We can obviously make it more accurate if you think thats best. I was just trying to make sure I could follow along easier with the previous behavior

src/app/components/feed/FeedImportDialog.js
```diff
@@ -114,7 +114,7 @@ const FeedImportDialog = ({
         open
       >
         <div className={dialogStyles['dialog-title']}>
-          <div>
+          <div className={dialogStyles['dialog-title-choice']}>
```
brianfleming, line 117: unrelated to the functionality in this PR, but as I was grabbing the styles to use the toggle buttons in the new merge dialog, figure I would just add this class so the styles/layout matches for both use cases

brianfleming:
Thanks for that push @caiosba reviewing now and I am seeing this state that I can't figure out:

Screenshot 2024-08-30 at 9 13 02 AM

I have output what the code thinks is current `{this.state.action}` at the top of the dialog which is `addThisToSimilar` which corresponds to the export option.The export is disabled, because I think this state is happening on media clusters with a published fact-check.

Import is the only thing available, which I think is correct, but its not selected. Can you take a look?

caiosba:
> Ah, sure @brianfleming ... I think it's just the logic of the initial state that needs to be adjusted... let me take a look.

Done, @brianfleming !

caiosba (approved):
Pair-programmed on this with Brian, so the pair review happened during development.

PR #2086, opened 2024-08-31
Should be able to create claims without fact-checks.

caiosba:
## Description

Disclaimer: I don't like the solution that I implemented here, tbh. But it was the simplest thing I could think of. Reminding the idea of the ticket (CV2-5134): do the least work possible that doesn't move us backward in order to (hopefully temporarily) support a single partner that still uses claims without fact-checks. So, I did here something that doesn't require any reversion in the backend, and that's actually what we did previously, after the release, to migrate claims without fact-checks to the new architecture. In a nutshell, create a fact-check with placeholder values, that are not indexed. Sub-tasks:

- [x] Pre-select default language in articles forms
- [x] Fix "unknown language" bug
- [x] These fields should not be required anymore: fact-check title, fact-check summary and language
- [x] Should be able to create claim without fact-check
- [x] If fact-check fields are not ready, the "Unpublished Report" button should be disabled and display a tooltip
- [x] In the articles cards, display claim title and context if there is no fact-check title and summary
  - [x] Articles list
  - [x] Applied article card in item page
  - [x] "Recent articles" list in item page

Reference: CV2-5134.

## Type of change

## How has this been tested?

Manually. See checklist in the description.

## Checklist

### DIFF
--- a/localization/react-intl/src/app/components/article/ArticleForm.json
+++ b/localization/react-intl/src/app/components/article/ArticleForm.json
@@ -55,6 +55,16 @@
     "defaultMessage": "Assign this Fact Check to a media item to be able to edit and publish a report"
   },
   {
+    "id": "articleForm.reportDesignerTooltipTwo",
+    "description": "Tooltip for the report designer button",
+    "defaultMessage": "Fact-Check Title, Summary, and Language are required in order to to publish a Fact-Check report"
+  },
+  {
+    "id": "articleForm.reportDesignerTooltipThree",
+    "description": "Tooltip for the report designer button",
+    "defaultMessage": "Go to Fact-Check report editor"
+  },
+  {
     "id": "articleForm.publishedReport",
     "description": "A label on a button that opens the report for this item. This displays if the report for this media item is currently in the 'Published' state.",
     "defaultMessage": "Published report"
--- a/src/app/components/article/ArticleForm.js
+++ b/src/app/components/article/ArticleForm.js
@@ -82,6 +82,8 @@ const ArticleForm = ({
     }
     if (articleType === 'fact-check' && claimDescription && articleTitle && summary && language) {
       setCanPublish(true);
+    } else if (articleType === 'fact-check') {
+      setCanPublish(false);
     }
   }, [articleTitle, summary, claimDescription, language]);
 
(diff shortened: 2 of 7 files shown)


brianfleming:
Just one additional clarification about your steps in the PR @caiosba for the last item:
> In the articles cards, display claim title if there is no fact-check

You probably were already going to do this, but just to make sure:
Also show the claim "additional context" in the card description area if available.

So the logic is:
If there is a fact-check show the `title` and `summary`, if not, try to show the `claim title` and `additional context`

caiosba:
Yes that's clear, but thanks for putting it written here, @brianfleming !

brianfleming (commented):
@caiosba just a quick 👻 check in on this and its looking good.
I think its probably local data, hopefully, but I am still seeing the unknown language in the dropdown. Again, not sure if this is something to be concerned about, since you cannot select it. Just looks confusing:

Screenshot 2024-09-02 at 11 08 17 AM

brianfleming (changes requested):
I did find an issue @caiosba 
Take a look at the clip below. If I start to add a fact-check, then back out. I can still publish.
Its not in the clip, but if I remove the content from those fields, then still hit the create and publish button, the "old" content is actually saved and will be on the report page

https://github.com/user-attachments/assets/14525b9b-9d98-4653-9927-28eb45d9377b

caiosba:
Thanks @brianfleming :ghost: 
I'm going to work on this last bug you reported. For the "unknown language", maybe it's an existing fact-check you had that actually had the language saved as such?

caiosba:
> Code LGTM. Suggestion: introduce a helper function for the repeated testing of "value exists and is not a - placeholder, otherwise use that"

Done!

caiosba:
I fixed the "create & publish" bug, @brianfleming , please review :ghost:

brianfleming (approved):
Thanks @caiosba we can get this into QA and let Jessie try to find any issues now I think.

I am going to approve, so I don't hold you up, afterall i am a 👻 
but one thing to fix before merging, is that when you return the "combo" tooltip, there is a missing space in `reportFact-Check Title` because you are combining phrases I believe

Screenshot 2024-09-02 at 12 52 35 PM

caiosba:
> Thanks @caiosba we can get this into QA and let Jessie try to find any issues now I think.
> 
> I am going to approve, so I don't hold you up, afterall i am a 👻 but one thing to fix before merging, is that when you return the "combo" tooltip, there is a missing space in `reportFact-Check Title` because you are combining phrases I believe
> 
> Screenshot 2024-09-02 at 12 52 35 PM

Ah, right! I just fixed it by making sure we only show one message in the tooltip each time.

PR #2107, opened 2024-09-13
Epic/CV2-4111: Workspace analytics

danielevalverde:
### Articles Dashboard
- [x] QueryRenderer
- [x] Basic page layout
- [x] Shared CSS module for both pages
  - [x] 2 column layout
- [x] Filter header area
- [x] Reusable filter components
  - [x] TimeFrame
  - [x] Languages
- [x] Add charts in each page with no data/empty states while GraphQL layer is being worked on
- [x] Connect GraphQL data to charts
- [x] Connect filter components to query

### Tiplines Dashboard
- [x] QueryRenderer
- [x] Basic page layout
- [x] Shared CSS module for both pages
  - [x] 2 column layout
- [x] Filter header area
- [x] Reusable filter components
  - [x] TimeFrame
  - [x] Languages
  - [x] Platforms
- [x] Add charts in each page with no data/empty states while GraphQL layer is being worked on
- [x] Connect GraphQL data to charts
- [x] Connect filter components to query

References: CV2-4936 CV2-4937

## Type of change

## How has this been tested?

Tested manually viewing each dashboard with random generated data and zero data.

- Etc.

## Checklist


src/app/components/dashboard/Dashboard.module.css
```diff
@@ -0,0 +1,54 @@
...
+
+    .dashboard-filter-area {
+      background: var(--color-beige-86);
+      border-radius: var(--border-radius-default);
+      display: flex;
+      gap: 16px;
+      padding: 16px;
+    }
+
+    .dashboard-two-column {
```
brianfleming, line 22: will start to review this @amoedoamorim just adding a quick comment that I think if you add `flex-wrap: wrap;` to this, then we will get some nice reponsiveness on the dashboard

src/app/components/dashboard/TimelineArticlesCreatedAndUpdated.js
```diff
@@ -0,0 +1,43 @@
...
+import { FormattedMessage } from 'react-intl';
+import { createFragmentContainer, graphql } from 'react-relay/compat';
+import TimelineWidget from '../cds/charts/TimelineWidget';
+
+const TimelineArticlesCreatedAndUpdated = ({ statistics }) => {
+  //
+  // TODO: Format tooltip to show both created and updated articles
+  //
+
+  console.log('TimelineArticlesCreatedAndUpdated statistics:', statistics.number_of_articles_created_by_date); // eslint-disable-line no-console
```
sarahkeh, line 12: Are these console statements still needed?

src/app/components/media/MediaTypeDisplayName.js
```diff
@@ -1,6 +1,73 @@
...
+    id: 'media.typeVideo',
+  },
+  UploadedAudio: {
+    defaultMessage: 'Audio',
+    description: 'Label to show that the type of media is an audio file',
+    id: 'media.typeAudio',
+  },
+  Blank: {
+    defaultMessage: 'Imported fact-check',
+    description: 'Label to show that the type of media is was imported into the application',
```
sarahkeh, line 33: small typo, 'is' should be removed

brianfleming (changes requested):
@amoedoamorim et al. Couple Changes found while reviewing and one question:
- Can we update the page title for dashboard to match the other patterns? As you can see in this screenshot the page title still references the All page after I go to the dashboard
Screenshot 2024-10-30 at 9 39 24 AM


- One situation that I missed during the review of the stacked bar is when one of the slices is 0 and it appears last. As you can see in the screenshot below, the visual looks a bit broken because the right side doesn't end up rounded. Really it should always have a rounded end since that the container, compare the two here, let me know if my poor explanation could be better
Screenshot 2024-10-30 at 9 37 25 AM

## Question
Not sure how much control we have over this via the viz package, but if you look at the screenshot below you can see that the timeline dates can get in weird spacing situations, anything we can do?
Screenshot 2024-10-30 at 9 38 00 AM

brianfleming:
Another small thing to address @amoedoamorim 
Article tags are "the same" as media tags, so this empty state can go to the tags settings just like media:

Screenshot 2024-10-30 at 9 56 11 AM

caiosba:
Tested locally and works well!

sarahkeh (approved):
I left a couple tiny comments, but looks good! The display components all follow the same pattern so it's easy to read.

amoedoamorim:
> Not sure how much control we have over this via the viz package, but if you look at the screenshot below you can see that the timeline dates can get in weird spacing situations, anything we can do?

@brianfleming yeah, that sucks. I found it seems to occur at random times but if you navigate away and back then the timeline looks evenly spaced with the same data. Since this looks like a weird occasional edge case, I'll save this for last as it will probably take more research, trial and error than the other fixes I need to get in.

brianfleming:
> > Not sure how much control we have over this via the viz package, but if you look at the screenshot below you can see that the timeline dates can get in weird spacing situations, anything we can do?
> 
> @brianfleming yeah, that sucks. I found it seems to occur at random times but if you navigate away and back then the timeline looks evenly spaced with the same data. Since this looks like a weird occasional edge case, I'll save this for last as it will probably take more research, trial and error than the other fixes I need to get in.

yeah thats what I figured @amoedoamorim 
agree, leave till the end and I don't think we consider it a release blocker either

brianfleming (approved):
great, thanks for those changes @amoedoamorim 
agreed, lets get this to QA and find out all the other things we need to do!

PR #2114, opened 2024-09-18
4933 – Analytics Widget: List

vasconsaurus:
## Description

This implements the list widget component. 
The list widget allows us to show “top lists” of data point in Check.

#### Highlights:
- Title is customizable
- Main color is customizable via CSS var
- List can contain links or text
    - Links may take users to item pages, or perhaps a list with a filter already applied
- List may contain counts
    - Our first uses will all have counts, but this should not be a requirement
- List should be able to handle any number of items
    - our first uses will be limited to 5, but this should not be a hard coded amount

#### Images
Screenshot 2024-09-18 at 09 23 07
Screenshot 2024-09-18 at 09 23 26

References: 4933

## Type of change

## How has this been tested?

I tested by adding it to the Sandbox and manually testing it there. And by adding two unit tests:

```
npm run test:unit ListWidget.test.js
npm run test:unit ListWidgetItem.test.js
```

- Etc.

## Checklist

### DIFF
--- a/src/app/components/cds/charts/ListWidget.js
+++ b/src/app/components/cds/charts/ListWidget.js
@@ -13,9 +13,10 @@ const ListWidget = ({
       {title}
     </div>
     <ul>
-      { Object.values(items).map(i => (
+      { items.map(i => (
         <ListWidgetItem
           itemCount={i.itemCount}
+          itemLink={i.itemLink}
           itemText={i.itemText}
           key={i.id}
         />
--- a/src/app/components/cds/charts/ListWidget.js
+++ b/src/app/components/cds/charts/ListWidget.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ListWidgetItem from './ListWidgetItem';
+
+const ListWidget = ({
+  color,
+  items,
+  title,
+}) => (
+  <div>
+    {color}
+    {title}
+    <ul>
+      { Object.values(items).map(i => (
+        <ListWidgetItem
+          itemCount={i.itemCount}
+          itemText={i.itemText}
+          key={i.id}
+        />
+      ))}
+    </ul>
+  </div>
+);
+
+ListWidget.defaultProps = {
+  color: 'var(--color-pink-93)',
+  items: null,
+  title: null,
+};
+
+ListWidget.propTypes = {
+  color: PropTypes.string,
+  items: PropTypes.object,
+  title: PropTypes.string,
+};
+
+export default ListWidget;
--- a/src/app/components/cds/charts/ListWidget.js
+++ b/src/app/components/cds/charts/ListWidget.js
@@ -1,15 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ListWidgetItem from './ListWidgetItem';
+import styles from './ListWidget.module.css';
 
 const ListWidget = ({
   color,
   items,
   title,
 }) => (
-  <div>
-    {color}
-    {title}
+  <div className={styles.listWidgetWrapper} style={{ backgroundColor: color }}>
+    <div className={styles.listWidgetTypography}>
+      {title}
+    </div>
     <ul>
       { Object.values(items).map(i => (
         <ListWidgetItem
--- a/src/app/components/cds/charts/ListWidget.js
+++ b/src/app/components/cds/charts/ListWidget.js
@@ -15,9 +15,9 @@ const ListWidget = ({
     <ul>
       { items.map(i => (
         <ListWidgetItem
-          itemCount={i.itemCount}
           itemLink={i.itemLink}
           itemText={i.itemText}
+          itemValue={i.itemValue}
           key={i.id}
         />
       ))}
--- a/src/app/components/cds/charts/ListWidget.js
+++ b/src/app/components/cds/charts/ListWidget.js
@@ -31,7 +31,7 @@ ListWidget.defaultProps = {
 
 ListWidget.propTypes = {
   color: PropTypes.string,
-  items: PropTypes.arrayOf(PropTypes.shape({ link: PropTypes.string, text: PropTypes.string, value: PropTypes.string })).isRequired,
+  items: PropTypes.arrayOf(PropTypes.shape({ itemLink: PropTypes.string, itemText: PropTypes.string, itemValue: PropTypes.string })).isRequired,
   title: PropTypes.node.isRequired,
 };
 
--- a/src/app/components/cds/charts/ListWidget.js
+++ b/src/app/components/cds/charts/ListWidget.js
@@ -31,7 +31,7 @@ ListWidget.defaultProps = {
 
 ListWidget.propTypes = {
   color: PropTypes.string,
-  items: PropTypes.array.isRequired,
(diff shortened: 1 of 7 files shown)


vasconsaurus:
Updated Sandbox:
Screenshot 2024-09-18 at 09 57 16

vasconsaurus:
front-end friends (@amoedoamorim @danielevalverde @sarahkeh) if any of you have time, could you take a look at this PR? Then I can get any fixes done by today 💪

src/app/components/cds/charts/ListWidgetItem.test.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  const nullProps = {
+    id: null,
+    itemLink: null,
+    itemText: null,
+    itemValue: null,
+  };
+
+  it('renders without crashing if props are null', () => {
```
amoedoamorim, line 24: suggestion: add a test case: "renders a link if itemLink prop contains an url"

src/app/components/cds/charts/ListWidgetItem.js
```diff
@@ -0,0 +1,36 @@
...
+
+const ListWidgetItem = ({
+  id,
+  itemLink,
+  itemText,
+  itemValue,
+}) => (
+  <li className={styles.listWidgetItemWrapper} key={id}>
+    {
+      itemLink ?
```
amoedoamorim, line 13: suggestion: validate url in itemLink

src/app/components/cds/charts/ListWidget.test.js
```diff
@@ -0,0 +1,54 @@
...
+describe('<ListWidget />', () => {
+  const props = {
+    color: 'var(--color-purple-92)',
+    items,
+    title: 'Title',
+  };
+
+  it('renders without crashing', () => {
+    mountWithIntl(<ListWidget {...props} />);
+  });
```
amoedoamorim, line 43: suggestion: assert that everything is rendered as expected, for each prop.
E.g.: ListWidgetItem are rendered as many as are there int the `items` prop.

src/app/components/cds/charts/ListWidgetItem.js
amoedoamorim: Don't worry too much about it, since we'll probably only have trusted backend generated links, not user input.
But it's a nice to have, if you got enough time for this.

src/app/components/cds/charts/ListWidget.js
```diff
@@ -0,0 +1,38 @@
...
+  </div>
+);
+
+ListWidget.defaultProps = {
+  color: 'var(--color-pink-93)',
+};
+
+ListWidget.propTypes = {
+  color: PropTypes.string,
+  items: PropTypes.array.isRequired,
```
amoedoamorim, line 34: Please specify items format: `PropTypes.arrayOf(PropTypes.shape({ ... })).isRequired`

amoedoamorim (changes requested):
Great work, Manu! My only actual requirement is to specify the array item shape. But if you manage to get the suggestions in, extra kudos for you! :smile:

danielevalverde (commented):
Other than Alex's point, the code looks great to me and the functionality works well locally :)

src/app/components/cds/charts/ListWidgetItem.js
vasconsaurus: @amoedoamorim questions:

I tried both using `URL.canParse` and the helper `validateURL` and ran into different problems with both:
- `URL.canParse`: it renders, when manually testing it works, whether I send null an empty string or a malformed URL. But when I run the tests it fails `Uncaught [TypeError: URL.canParse is not a function]`
-  `validateURL`: it fails both when I test manually and run tests, with `Error: Uncaught [TypeError: Cannot read property 'length' of null]`

What the code looked like:
```
    {
      URL.canParse(itemLink) ?
        <div><a href={itemLink}>{itemText}</a></div>
        :
        <div className={styles.listWidgetItemText}>{itemText}</div>
    }
```

```
    {
      validateURL(itemLink) ?
        <div><a href={itemLink}>{itemText}</a></div>
        :
        <div className={styles.listWidgetItemText}>{itemText}</div>
    }
```

amoedoamorim: I never user `URL.canParse`, I'm actually learning about it right now! :smile:
It seems it's a new Web API that's not implemented by the Node version that runs the unit tests, this is why it works on the browser and fails on the test.

I think `validateURL` expects a string as an argument. We can fix the helper function to return false if argument is not a string.

vasconsaurus: > It seems it's a new Web API that's not implemented by the Node version that runs the unit tests, this is why it works on the browser and fails on the test.

Aaaaaaaaaaa! That makes sense! 😅 

> I think `validateURL` expects a string as an argument. We can fix the helper function to return false if argument is not a string.

That fixed it! But I thought it was better to create `isValidURL` that just returns a boolean? It runs `linkify.test(value)` instead of `match`. But let me know if you prefer that I change `validateURL`

brianfleming (changes requested):
@vasconsaurus exciting to see this in progress, its looking great! I have a couple of visual changes that I think we should make:

1. Its currently hard to "scan" the list since the text could end in different places. If you look at the screenshot below. the last item's text goes "under" the number above, so if you are just looking at the list, you eye has to jump around. My suggestion is: Have the number area be a fixed width that can accept up to 7 digits, I think this is the max number we will probably see in the system. If you do it this way, then the text column will always "end" at the same place, so you it will be easier to visually run the list.
Screenshot 2024-09-19 at 9 30 05 AM

2. If you make the number column a fixed width, you will avoid the situation you see in the screenshot below, where when the widget gets very small the number column may get cut off and you will possibly not see what the real number is. So what this means is that the number should always be fully visible, but the text column can get as small as possible
Screenshot 2024-09-19 at 9 30 45 AM

Let me know if thats confusing or if I can add a better explanation.

vasconsaurus:
I thought I left a note here but it was the other PR 😅 @brianfleming @amoedoamorim can you please review again and let me know if any changes are needed?

src/app/components/cds/charts/ListWidget.js
```diff
@@ -0,0 +1,38 @@
...
+  </div>
+);
+
+ListWidget.defaultProps = {
+  color: 'var(--color-pink-93)',
+};
+
+ListWidget.propTypes = {
+  color: PropTypes.string,
+  items: PropTypes.arrayOf(PropTypes.shape({ link: PropTypes.string, text: PropTypes.string, value: PropTypes.string })).isRequired,
```
amoedoamorim, line 34: Hi @vasconsaurus, the propTypes shape doesn't match the actual data: `itemLink`, `itemText` and `itemValue`

amoedoamorim (dismissed):
:lgtm:

Awesome stuff, and thanks for the improvements, Manu! :tada:

src/app/components/cds/charts/ListWidget.js
amoedoamorim: Please update this and it should be good to go from my perspective! :smile:

vasconsaurus: 🤦‍♀️ sorry! updated!

amoedoamorim (approved):
:lgtm:

brianfleming (approved):
very nice @vasconsaurus ! learn to love the list

list

PR #2116, opened 2024-09-19
4934 – Analytics Widget: Number – Make sure missing elements don't affect the DOM

vasconsaurus:
## Description

For optional elements, its typically better to not have them influence the DOM if they are not present, since it can alter the layout.

Before when we left out eft out the contextText we could still see the <div.. that outputs it is still written to the screen. You can’t see anything visually, but what it does is make the widget artificially taller than it should be since it still takes up space.

When we don't add the div the padding of the widget is “correct”

References: 4934

#### Reference

Screenshot 2024-09-19 at 14 23 43

#### Fixed : DIV is not added when there is not `contextText`

Screenshot 2024-09-19 at 14 24 37

## Type of change

## How has this been tested?

- Etc.

## Checklist

### DIFF
--- a/src/app/components/cds/charts/NumberWidget.js
+++ b/src/app/components/cds/charts/NumberWidget.js
@@ -13,14 +13,9 @@ const NumberWidget = ({
     <div className={styles.numberWidgetTitle}>
       {title}
     </div>
-    { itemCount && itemCount !== '-' &&
+    { itemCount &&
       <div className={styles.numberWidgetItemCount}>
-        {itemCount} { unit && <span className={styles.numberWidgetUnit}>{unit}</span> }
-      </div>
-    }
-    { itemCount === '-' &&
-      <div className={styles.numberWidgetItemCount}>
-        {itemCount}
+        {itemCount} { unit && itemCount !== '-' && <span className={styles.numberWidgetUnit}>{unit}</span> }
       </div>
     }
     { contextText &&
--- a/src/app/components/cds/charts/NumberWidget.js
+++ b/src/app/components/cds/charts/NumberWidget.js
@@ -13,9 +13,16 @@ const NumberWidget = ({
     <div className="typography-subtitle2">
       {title}
     </div>
-    <div className={styles.numberWidgetItemCount}>
-      {itemCount} <span className={styles.numberWidgetUnit}>{!itemCount || itemCount === '-' ? null : unit} </span>
-    </div>
+    { itemCount && itemCount !== '-' &&
+      <div className={styles.numberWidgetItemCount}>
+        {itemCount} { unit && <span className={styles.numberWidgetUnit}>{unit}</span> }
+      </div>
+    }
+    { itemCount === '-' &&
+      <div className={styles.numberWidgetItemCount}>
+        {itemCount}
+      </div>
+    }
     { contextText &&
       <div className={styles.numberWidgetContextText}>
         {contextText}
--- a/src/app/components/cds/charts/NumberWidget.js
+++ b/src/app/components/cds/charts/NumberWidget.js
@@ -16,9 +16,11 @@ const NumberWidget = ({
     <div className={styles.numberWidgetItemCount}>
       {itemCount} <span className={styles.numberWidgetUnit}>{!itemCount || itemCount === '-' ? null : unit} </span>
     </div>
-    <div className={styles.numberWidgetContextText}>
-      {contextText}
-    </div>
+    { contextText &&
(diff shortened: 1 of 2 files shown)


brianfleming (changes requested):
nice @vasconsaurus 
I would just suggest to do the same with the unit. lets of an impact on layout since it appears in a horizontal lockup, but still worth making consistent

brianfleming (commented):
Thanks @vasconsaurus this does accomplish the goal. However, maybe the other engineers should do a quick look to comment on if it might be clearer to code readers to instead do the `numberWidgetItemCount` this general way:

```
<div className={styles.numberWidgetItemCount}>
    {itemCount} { unit && <span className={styles.numberWidgetUnit}>{unit}</span> }
</div>
```

After all, we are "always" going to have a count (it just might be `-` sometimes), its just the unit that may or may not be there (null)

src/app/components/cds/charts/NumberWidget.js
```diff
@@ -13,12 +13,21 @@ const NumberWidget = ({
     <div className="typography-subtitle2">
```
brianfleming, line 13: oh also @vasconsaurus unrelated to this PR, but I noticed that in the js file here, you are using the typography class to set the styles, but in the css file for this, you also have `.numberWidgetTitle` which is unused. So you probably either want to use that className or remove it from the css file

vasconsaurus:
> Thanks @vasconsaurus this does accomplish the goal. However, maybe the other engineers should do a quick look to comment on if it might be clearer to code readers to instead do the `numberWidgetItemCount` this general way:
> 
> ```
> <div className={styles.numberWidgetItemCount}>
>     {itemCount} { unit && <span className={styles.numberWidgetUnit}>{unit}</span> }
> </div>
> ```
> 
> After all, we are "always" going to have a count (it just might be `-` sometimes), its just the unit that may or may not be there (null)

The problem I ran into with that one is if `itemCount` is `-`, then it will add the unit. So we get `- unit`. Is that behavior ok?

src/app/components/cds/charts/NumberWidget.js
vasconsaurus: Missed this! Thank you!

brianfleming:
> > Thanks @vasconsaurus this does accomplish the goal. However, maybe the other engineers should do a quick look to comment on if it might be clearer to code readers to instead do the `numberWidgetItemCount` this general way:
> > ```
> > <div className={styles.numberWidgetItemCount}>
> >     {itemCount} { unit && <span className={styles.numberWidgetUnit}>{unit}</span> }
> > </div>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After all, we are "always" going to have a count (it just might be `-` sometimes), its just the unit that may or may not be there (null)
> 
> The problem I ran into with that one is if `itemCount` is `-`, then it will add the unit. So we get `- unit`. Is that behavior ok?

ah, good point, I didn't think of that. you're right, when the count is `-` its better to not have the unit

vasconsaurus:
@amoedoamorim, when you have some time, could you do one final (~hopefully~ 😅 ) review? I made some changes after Brian's review.

src/app/components/cds/charts/NumberWidget.js
```diff
@@ -10,15 +10,24 @@ const NumberWidget = ({
...
+    <div className={styles.numberWidgetTitle}>
       {title}
     </div>
-    <div className={styles.numberWidgetItemCount}>
-      {itemCount} <span className={styles.numberWidgetUnit}>{!itemCount || itemCount === '-' ? null : unit} </span>
-    </div>
-    <div className={styles.numberWidgetContextText}>
-      {contextText}
-    </div>
+    { itemCount && itemCount !== '-' &&
```
sarahkeh, line 16: might be able to condense this to 
```
{ itemCount &&
      <div className={styles.numberWidgetItemCount}>
        {itemCount} { unit && itemCount !== '-' && <span className={styles.numberWidgetUnit}>{unit}</span> }
      </div>
    }
{ contextText &&
      <div className={styles.numberWidgetContextText}>
        {contextText}
      </div>
    }
```
Otherwise looks good!

vasconsaurus:
@brianfleming updated with @sarahkeh suggestion

brianfleming (approved):
niiiice @vasconsaurus ! Thanks for all the changes and improvements!

number-one

amoedoamorim (approved):
:lgtm:

PR #2127, opened 2024-09-24
Switch UX and Font Improvement

brianfleming:
## Description

Minor PR here, while working on some other features I noticed that the Switch control was using a different font than the rest of the Check interface. This is because the switch component uses some additional material ui components, notably the `FormControlLabel` which simply outputs a label to make the switch clickable.

In order to get the same font face as the rest of the UI, I chose to remove the dependency on `FormControlLabel`, so this PR:
- Makes the label for the switch be plain html
- in the sandbox instead of choosing where to output the label I changed it to just show all 4 variants, since its only 4 and its easier to see any layout drift
- do the necessary and sort prop types

## Type of change

## How has this been tested?

sandbox verification

Not much, there should be no visual differences other than the font-face

### AFTER - sandbox
Screenshot 2024-09-24 at 4 27 03 PM

### AFTER  - another variant with disabled and help shown
Screenshot 2024-09-24 at 4 31 47 PM

## Checklist


amoedoamorim (approved):
https://cdn.dribbble.com/users/577205/screenshots/2320187/media/306d8c851efeeef063211e48f6f0da4e.gif

PR #2142, opened 2024-10-01
Feature/CV2-5319-trash-relay-article-refresh

sarahkeh:
## Description

Page no longer needs a full reload after an item is trashed.  After trashing an item, a refresh=true string is pushed to the url, triggering a refresh of the current viewed list.  In the Update function, counters used for the sidebar are returned which update the values to reflect the moved article.  This was done by making every counter explicitly named, removing the middle ArticleCoreListCounter component, and moving the query for those values to the parent ArticleDrawer queryRenderer.  Also renamed ProjectsListCounter to DrawerListCounter to indicate a more generic use.  

References: CV2-5319

## Type of change

## How has this been tested?

Trashed and restored Fact Checks, Explainers, and Published fact-checks.  

If counters and lists aren't updating, check that you're in Production mode.   

## Checklist


src/app/components/drawer/DrawerArticles.js
```diff
@@ -59,7 +59,7 @@ const DrawerArticlesComponent = ({ team }) => {
               <div className={styles.listLabel}>
                 <FormattedMessage defaultMessage="Claim & Fact-Checks" description="Label for a list displayed on the left sidebar that includes items that have claim & fact-checks" id="articlesComponent.claimAndFactChecks" tagName="span" />
               </div>
-              <ArticleCoreListCounter teamSlug={team.slug} type="fact-check" />
+              <DrawerListCounter key={team.factChecksCount} numberOfItems={team.factChecksCount} />
```
amoedoamorim, line 62: These `key` props passed to the `DrawerListCounter` seem unnecessary as this isn't the output of a list.map and `DrawerListCounter` has no internal state that would need being reset upon key updating. Can you plz remove them?

src/app/components/article/ArticleTrash.js
```diff
@@ -20,6 +21,11 @@ const updateExplainer = graphql`
         id
         trashed
       }
+      team {
+        articles_count
```
amoedoamorim, line 25: Is this generic `articles_count` still needed now that we have aliases for each list?

amoedoamorim (approved):
Please review my comments and apply them if they're pertinent, other than that. Looks pretty good, approving.

src/app/components/article/ArticleTrash.js
sarahkeh: No, you're right, removed.

src/app/components/drawer/DrawerArticles.js
sarahkeh: Removed

PR #2162, opened 2024-10-11
allow webpage titles to wrap when they are unbroken strings [CV2-5460]

brianfleming:
## Description

Noticed while reviewing designs with internal folks that for webpage media cards, if the title happens to be a long unbroken string such as a URL, then in Chrome only, it causes overflow issues. There is also a less than optimal layout of errors in the cards. This PR addresses both. Also
- removes one unnecessary file, i felt this was unnecessary optimization to have the code separate when its only used in one place
- replaces material box with css

References: CV2-5460

## Type of change

## How has this been tested?

manual verification

Issue can be seen on QA here:
https://qa.checkmedia.org/check-capi-demo/media/38888

### AFTER TITLE
Screenshot 2024-10-11 at 11 26 16 AM

### AFTER Error
Screenshot 2024-10-11 at 11 26 42 AM

## Checklist


sarahkeh (approved):
much nicer, looks great!

PR #2165, opened 2024-10-15
Feature/CV2-5204 - Display article card always when list is returning, articles

sarahkeh:
## Description

Replaced SmallMediaCard with ArticleCard. By passing a fact-check or media value down to the search component, the component can determine which card type to render. Small optional props were added to the ArticleCard for this: a classNames prop in order to pass it the Selected styling, and a readOnly boolean for the tags.  From the search, I added several fact_check and claim_description fields that were needed for the Card, as well as the team's verification_statuses. 

I set the Tags component to ReadOnly as it didn't seem like an appropriate time for the user to make tag changes to a Fact-Check.  It also would involve mixing Relay and Fetch in one component which felt not great. 

Screenshot 2024-10-15 at 1 39 30 AM

References: CV2-5204

## Type of change

- [] New feature (non-breaking change which adds functionality)

## How has this been tested?

manually, checking both that the media cards in the other tabs were not impacted and the article card matched the card in the fact-checks list

Tags as Read-Only - this felt like the simpler user decision but I can change this easily if needed

## Checklist

### DIFF
--- a/src/app/components/cds/menus-lists-dialogs/TagPicker.js
+++ b/src/app/components/cds/menus-lists-dialogs/TagPicker.js
@@ -124,29 +124,31 @@ const TagPicker = ({
           )}
         </FormattedMessage>
       </Popover>
-      <Tooltip
-        arrow
-        placement="top"
-        title={
-          <FormattedMessage
-            defaultMessage="Manage Tags"
-            description="Tooltip message displayed on a tag item to let the user know they can manage the tags in the list"
-            id="taglist.tooltipManage"
-          />
-        }
-      >
-        <div>
-          <ButtonMain
-            className="int-tag-list__button--manage"
-            disabled={readOnly}
-            iconCenter={<TagMoreIcon />}
-            size="small"
-            theme="lightText"
-            variant="contained"
-            onClick={handleOpenMenu}
-          />
-        </div>
-      </Tooltip>
+      { !readOnly &&
+        <Tooltip
+          arrow
+          placement="top"
+          title={
+            <FormattedMessage
+              defaultMessage="Manage Tags"
+              description="Tooltip message displayed on a tag item to let the user know they can manage the tags in the list"
+              id="taglist.tooltipManage"
+            />
+          }
+        >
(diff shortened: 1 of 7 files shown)


brianfleming (commented):
@sarahkeh this is looking good. I will let others comment on the tech for actual approvals.
But I had one request, now that I am seeing this in real life, something to add to this PR that is kind of related.
Overtime the Tags List has changed, but I think we should make one more change to have this interaction a bit more clear:
- When the tag list is read only (like in this use case) can we hide the "manage tags" button instead of just disabling?
- I don't think there is value in seeing that button if you can't interact with it.

thoughts?

danielevalverde (commented):
:lgtm:

sarahkeh:
> @sarahkeh this is looking good. I will let others comment on the tech for actual approvals. But I had one request, now that I am seeing this in real life, something to add to this PR that is kind of related. Overtime the Tags List has changed, but I think we should make one more change to have this interaction a bit more clear:
> 
> * When the tag list is read only (like in this use case) can we hide the "manage tags" button instead of just disabling?
> * I don't think there is value in seeing that button if you can't interact with it.
> 
> thoughts?

I had that thought, too.  Just pushed the change, makes more sense than an unused button 

Screenshot 2024-10-15 at 10 58 25 AM

src/app/components/cds/menus-lists-dialogs/TagPicker.js
```diff
@@ -124,29 +124,31 @@ const TagPicker = ({
...
+              defaultMessage="Manage Tags"
+              description="Tooltip message displayed on a tag item to let the user know they can manage the tags in the list"
+              id="taglist.tooltipManage"
+            />
+          }
+        >
+          <div>
+            <ButtonMain
+              className="int-tag-list__button--manage"
+              disabled={readOnly}
```
amoedoamorim, line 142: `disabled={readOnly}` should be removed now that the button isn't rendered

amoedoamorim (approved):

PR #2177, opened 2024-10-31
cv2-3244 New Tab Component

sarahkeh:
- [x] Tab component - use ButtonMain as guideline but not base
  - [x] Size prop: default and large (might make more sense to put this prop on the Tab Wrapper)
  - [x] Icon prop: none, left, right, center (hide label for center but need label for Select dropdown)
  - [x] State prop: active, inactive
  - [x] Disabled prop
  - [ ] Hover styling for active, inactive
- [x] Tab Wrapper - 
  - [x] Variant prop: default or banner 
  - [x] Deep linking - value of selected tab put into tab parameter in url, url value should override default selected
- [x] Responsiveness - When tabs would otherwise overflow the container, switch to Select component.  
  - [x] Tab component returns select option
  - [x] TabWrapper returns Select component
- [x] Add to Sandbox
- [x] Add to Media Cluster item page
- [x] Unit test for correct rendering of tabs 

## Description

Marking as ready even though the integration tests are failing, I'll fix those on Tuesday. (might need to pair on that, the integration tests are run in a screen that causes the Select to be used and it doesn't seem to be finding the right <option>)

Replaces MUI Tabs and Tab components with custom cds TabWrapper and Tab.  TabWrapper takes in an object array that then is used to render either Select or a list of Tabs. 

References: CV2-3244

## Type of change

## How has this been tested?

Unit test, manually via the Sandbox and the Media component

Determining the correct width and when to switch between Tabs and Select was a journey but the solution I found was pretty interesting. 

## Checklist

- [x] If my components involve user interaction - specifically button, text fields, or other inputs - I have added a [BEM-like class name](https://meedan.atlassian.net/wiki/spaces/ENG/pages/1327628289/Naming+conventions+for+interactive+elements) to the element that is interacted with


brianfleming (changes requested):
@sarahkeh nice progress this is looking good, the responsive change to select is great.
Couple things to fix/check on:
- I don't think the deep linking is working on the first use of the tabs on the media cluster page, is that still WIP or is there something broken?
- When the tabs are set to iconCenter can you adjust the padding so the tab themselves end up squares? Take a look at the button main component and you will see something similar. When there is text the padding is different then when its just an icon

Screenshot 2024-11-11 at 9 07 22 AM


- There is something weird going on with the padding on the media cluster item page, take a look at the attached video and you will see it change for some reason, but just on the requests:

https://github.com/user-attachments/assets/7d7eff84-d8a9-45dd-be81-74c252dd8f20

brianfleming:
@sarahkeh I noticed one additional very minor thing. there is a tiny point when resizing before the `<Select...` kicks in where depending on the content of the tabs they can wrap. I think we can get around this just by not allowing the text in the tab to wrap:

Screenshot 2024-11-12 at 8 46 57 AM

sarahkeh:
@brianfleming Made the listed changes you pointed out, and fixed the deep linking (it would pull a tab parameter if it was already in the url but wasn't putting any up there)

brianfleming (changes requested):
Changes looking great @sarahkeh couple things:
- FYI: i just pushed up a different way to do bolding for these tabs so the sizing doesn't jump around, was easier for me just to do it then ask for that small change.

However....I did find two things to address

1. should clean up this new warning based on the PR checklist:

Screenshot 2024-11-13 at 10 54 16 AM

2. I'm trying to figure out how to replicate this, so far I think it "might" be going to a media page with the width of the window set to show the `<Select...` anyway, it throws this UI error and console:

Screenshot 2024-11-13 at 10 54 31 AM

brianfleming:
forgot to add this about that error @sarahkeh 
going to the media item page, with a small window AND the item has suggestions
^ I think the suggestions part is important, but having trouble confirming that

brianfleming:
yeah, I think thats the case with the suggestions @sarahkeh take a look at this recording of me loading the screen, resizing then refreshing:


https://github.com/user-attachments/assets/a760cf25-4d3e-41bb-aaea-5346341d85fd

sarahkeh:
@brianfleming Thanks for the recording, that helped me narrow it down a lot! It was infinitely resizing so I pulled out the setting of the fullTabWidth from the resize check and it loads correctly now. The other issues should be fixed as well

amoedoamorim (approved):
LGTM

PR #2200, opened 2024-11-13
Fix Error when searching for removed imported fact-check 

danielevalverde:
## Description

Add safe navigation to avoid the crash with the error:

`AutoCompleteMediaItem.js:258 TypeError: Cannot read properties of null (reading 'updated_at')
    at AutoCompleteMediaItem.js:371:51
    at Array.map (<anonymous>)
    at bq (AutoCompleteMediaItem.js:333:32)
    at to (react-dom.production.min.js:157:137)
    at Yo (react-dom.production.min.js:180:154)
    at Hs (react-dom.production.min.js:269:343)
    at Mu (react-dom.production.min.js:250:347)
    at wu (react-dom.production.min.js:250:278)
    at _u (react-dom.production.min.js:250:138)
    at fu (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at Ha (react-dom.production.min.js:122:325)
    at Ua (react-dom.production.min.js:123:61)
    at Wa (react-dom.production.min.js:122:428)
    at su (react-dom.production.min.js:237:203)
    at xo (react-dom.production.min.js:170:408)`

that was happening when searching for removed imported fact-check, probably due to a race condition. 

Reference: CV2-5600

## Type of change

## Checklist

### DIFF
--- a/src/app/components/media/AutoCompleteMediaItem.js
+++ b/src/app/components/media/AutoCompleteMediaItem.js
@@ -388,9 +388,9 @@ const AutoCompleteMediaItem = (props, context) => {
                     statusColor={currentStatus ? currentStatus.style?.color : null}
                     statusLabel={currentStatus ? currentStatus.label : null}
                     summary={isFactCheckValueBlank(projectMedia.fact_check.summary) ? projectMedia.fact_check.claim_description?.description : projectMedia.fact_check.summary}
-                    tags={projectMedia.fact_check?.tags}
+                    tags={projectMedia.fact_check.tags}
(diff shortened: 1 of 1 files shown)


amoedoamorim (changes requested):
Just did a paired review with Dani and we're going to skip returning the `ArticleCard` entirely if `projectMedia.fact_check` is null

amoedoamorim (approved):
:lgtm:

PR #2215, opened 2024-11-25
Bot preview tweaks

brianfleming:
## Description

While reviewing the progress on the bot preview I have made some adjustments to styles and layout because things just feel different when you go from design to real life. Some highlights of this PR:

- Added slight transition animation to the device toggle because 🤷 why not?
- Adjusted the layout a bit so it better matches the figma layout, which is not super obvious how to approach when looking at the designs, this is so we don't get into a weird layout situation later when we add more functionality
- tooltip for the device toggle
- fixed som icon and font sizes
- the font size for the device preview was set too high up in the cascade so it was making all fonts below it too large
- rename some welcome area css because it won't always be a welcome message in the future...premature optimization really

References: CV2-5571

## Type of change

## How has this been tested?

manual verification

not much, just style changes

## Checklist

### DIFF
--- a/localization/react-intl/src/app/components/cds/mockups/DeviceMockupComponent.json
+++ b/localization/react-intl/src/app/components/cds/mockups/DeviceMockupComponent.json
@@ -1,0 +1,7 @@
+[
+  {
+    "id": "deviceMockup.devicesTooltip",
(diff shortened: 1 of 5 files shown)


brianfleming:
reviewers, not sure what the failure message on the test is, appears on the epic PR as well

amoedoamorim (approved):

PR #2216, opened 2024-11-27
remove unnecessary code causing content warning broken state

brianfleming:
## Description

@amoedoamorim I noticed a bug on QA from my last PR:
https://github.com/meedan/check-web/pull/2214

It was causing the content warning to be clipped, in cases like when the media is no longer available, or being block by the browser (like mine was doing) see the screenshots below. This is just a quick css fix for that case

Screenshot 2024-11-27 at 2 09 58 PM
### In this case:
Screenshot 2024-11-27 at 2 09 51 PM

### DIFF
--- a/src/app/components/media/MediaCardLarge.module.css
+++ b/src/app/components/media/MediaCardLarge.module.css
@@ -9,6 +9,7 @@
   > .video-media-card,
   > .webpage-media-card {
     flex: 1 1;
+    min-height: calc(226px * (16/9));
     min-width: calc(200px * (16/9));
   }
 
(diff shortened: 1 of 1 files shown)


src/app/components/media/MediaCardLarge.module.css
```diff
@@ -9,13 +9,10 @@
   > .video-media-card,
   > .webpage-media-card {
     flex: 1 1;
+    min-height: calc(226px * (16/9));
     min-width: calc(200px * (16/9));
   }
 
-  .aspect-ratio-media-card {
```
brianfleming, line 15: removing this is the actual fix. The min-height above is just to be extra careful

brianfleming:
@amoedoamorim just a quick PR FYI for you since you reviewed the PR that I accidentally caused this issue in

amoedoamorim (approved):

PR #2222, opened 2024-12-04
Fix/cv2 5735 apply admin filter to all thumbnails

sarahkeh:
## Description

The admin content filter now puts a value to local storage.  This lets items not in the component tree pull from the same value.  Added a check in the ItemThumbnail component that looks for this flag and uses that to also determine if content should be hidden.  Also added a prop to ItemThumbnail that by default ignores this new localstorage value, to avoid impacts to other pages in the application.  For the item main icon and the search results in the Merge items, that value is set to false, allowing the thumbnail to be hidden by the admin flag.  

References: CV2-5735

## Type of change

## How has this been tested?

Manually

- Etc.

## Checklist

### DIFF
--- a/src/app/components/cds/media-cards/ItemThumbnail.js
+++ b/src/app/components/cds/media-cards/ItemThumbnail.js
@@ -9,20 +9,20 @@ import Tooltip from '../../cds/alerts-and-prompts/Tooltip';
 import styles from './ItemThumbnail.module.css';
 
 const ItemThumbnail = ({
+  ignoreGeneralContentMask,
   maskContent,
   picture,
   size,
   type,
   url,
 }) => {
-  const [contentMask, setContentMask] = React.useState(window.storage.getValue('contentMask') === 'true');
-  const mediaType = type === 'Link' ? mediaTypeFromUrl(url) : type;
-  const isHidden = contentMask || maskContent;
+  const [generalContentMask, setGeneralContentMask] = React.useState(window.storage.getValue('contentMask') || maskContent);
+  const isHidden = maskContent || (!ignoreGeneralContentMask && generalContentMask === 'true');
 
   useEffect(() => {
     const handleMaskChange = (event) => {
       if (event.key === 'contentMask') {
-        setContentMask(event.newValue === 'true');
+        setGeneralContentMask(event.newValue);
       }
     };
 
@@ -32,14 +32,6 @@ const ItemThumbnail = ({
       window.removeEventListener('storage', handleMaskChange);
     };
   }, []);
-
-  useEffect(() => {
-    if (contentMask === 'true') {
-      console.log('isHidden true', contentMask); //eslint-disable-line
-    } else {
-      console.log('isHidden false', contentMask); //eslint-disable-line
-    }
-  }, [contentMask]);
 
   if (!type && !picture) {
     return (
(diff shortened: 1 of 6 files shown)


brianfleming (commented):
@sarahkeh a bit hard to explain, and also not sure if what I am seeing is part of this work, but its at the very least thematically aligned:

There is some sort of bug/deoptimization around the thumbnail in the header and using the admin protections and using the content warning. It can get in a state where the header thumbnail doesn't always "respect" the user action. Take a look at the screen recording and see if this helps, as I said, tough to explain:

https://github.com/user-attachments/assets/114c19ec-50a9-46f7-889d-2283fb623a16

let me know if i can provide anymore info or what your thoughts are if this is aligned. It also might be something to solve "later" in the next steps of the epic work 🤷

sarahkeh:
@brianfleming I think this might be a separate issue.... I played around in QA and setting the content warning updates the thumbnail because it's a data update, but when it's temporarily lifted, that's just a temporary front end tag and, like how the Admin filter used to be, it's in a different component tree.  So it's similar, but unlike the admin/general content filter in the epic, it's very item-specific.... Personally it makes sense that the "Temporarily view content" only shows the large preview, since that's where the button is.  But maybe adding a "peek" button or something on the thumbnails might be helpful?

brianfleming:
> @brianfleming I think this might be a separate issue.... I played around in QA and setting the content warning updates the thumbnail because it's a data update, but when it's temporarily lifted, that's just a temporary front end tag and, like how the Admin filter used to be, it's in a different component tree. So it's similar, but unlike the admin/general content filter in the epic, it's very item-specific.... Personally it makes sense that the "Temporarily view content" only shows the large preview, since that's where the button is. But maybe adding a "peek" button or something on the thumbnails might be helpful?

I think some of that makes sense @sarahkeh but there is something that I unfortunately didn't show in the video, but here is a video a followup question:

why is it that the header thumbnail is the only one that sometimes I can't get to uncover. take a look at this vid, all all the media responding to my actions except the main one:

https://github.com/user-attachments/assets/7b940fb8-0d81-40ea-826a-6e6764853752

amoedoamorim:
@brianfleming I had a 1:1 with Sarah today and I see there are some problems with this functionality architecture.

As you know these two forms of hiding content are working independently, the per item content flag is data-driven and the admin screen is a frontend-only state.

The problem evidenced in the first video is because the components are under different Relay containers, unfortunately.
Because of how the code evolverd over time, `ItemThumbnail` in under a Relay classic container at `MediaActionsBar` and the `MediaCardLarge` is under a Relay modern QueryRenderer in `Media.js`. For this reason when you add content flag to an item, the large preview is hidden immediately but the thumbnail isn't aware of the change. But if you reload the page both get the updated data with the content flag set and will be hidden.
 
And the problem you evidence in the second video is probably because both flags function independently and the view temporarily toggle is local to `AspectRatio` which in effect applies the content cover. It seem that the item content flag is active, but the larger preview had the `Temporarily view content` on but that doesn't propagate to the item thumbnail. But you were clicking the Pause/resume admin screen button that also wouldn't affect the item content flag.

I think some refactoring there so the whole item page has a single QueryRenderer would make sense as we need to get rid of the Relay Classic stuff and would simplify things a great deal. We should also flatten `MediaPage`, `MediaPageLayout`, `Media` and maybe even `MediaComponent`. Too many layers there unnecessarily, imo.

So another benefit of a refactoring here is to lift that temporary view state management to the common ancestor  `MediaComponent` which would propagate the item content flag data to all its children.

I'd love to chat and structure a coherent architecture that can stand future iterations rather than just patching this already bad scenario.

brianfleming:
> @brianfleming I had a 1:1 with Sarah today and I see there are some problems with this functionality architecture.
> 
> As you know these two forms of hiding content are working independently, the per item content flag is data-driven and the admin screen is a frontend-only state.
> 
> The problem evidenced in the first video is because the components are under different Relay containers, unfortunately. Because of how the code evolverd over time, `ItemThumbnail` in under a Relay classic container at `MediaActionsBar` and the `MediaCardLarge` is under a Relay modern QueryRenderer in `Media.js`. For this reason when you add content flag to an item, the large preview is hidden immediately but the thumbnail isn't aware of the change. But if you reload the page both get the updated data with the content flag set and will be hidden.
> 
> And the problem you evidence in the second video is probably because both flags function independently and the view temporarily toggle is local to `AspectRatio` which in effect applies the content cover. It seem that the item content flag is active, but the larger preview had the `Temporarily view content` on but that doesn't propagate to the item thumbnail. But you were clicking the Pause/resume admin screen button that also wouldn't affect the item content flag.
> 
> I think some refactoring there so the whole item page has a single QueryRenderer would make sense as we need to get rid of the Relay Classic stuff and would simplify things a great deal. We should also flatten `MediaPage`, `MediaPageLayout`, `Media` and maybe even `MediaComponent`. Too many layers there unnecessarily, imo.
> 
> So another benefit of a refactoring here is to lift that temporary view state management to the common ancestor `MediaComponent` which would propagate the item content flag data to all its children.
> 
> I'd love to chat and structure a coherent architecture that can stand future iterations rather than just patching this already bad scenario.

@amoedoamorim and @sarahkeh got it!
thanks for the comprehensive look into all of this. I agree with the conclusions that we need a larger refactor and I am not surprised its a Relay issue!

Since a lot of this is pre-existing and we are only really "seeing" some of these edge cases because of improvements @sarahkeh is making. I say we just keep moving forward with whatever path you both think is best, knowing that we are moving in the right direction. 

We can chat about the bigger refactor for a coherent architecture whenever makes sense, I think we are getting to a view of Check that we can be confident in a good choice for that refactor.

amoedoamorim:
Hi @sarahkeh ! Code looks good, but I'm failing to understand why we need the new `ignoreGeneralContentMask` prop. Isn't the mask supposed to be applied everywhere, when active?

sarahkeh:
@amoedoamorim Eventually, yes, but for right now it's still contained just in the Media Item page.  The change that moves the content filter to the List view is a later ticket

amoedoamorim (approved):

PR #2232, opened 2024-12-17
Added orientation and theme prop to TogglebuttonGroup

sarahkeh:
## Description

Added two new props to the ToggleButtonGroup component: theme and orientation.  The new theme is called 'setting' to avoid overlapping with other theme names.  The orientation defaults to horizontal but can be toggled to vertical.  I think our version of the MaterialUI toggle button group predates the orientation prop (it's still in the /lab/ library, so likely still in development), so I had to add a custom settings for it.  

All new props can be tested in the Sandbox

References: CV2-5734

## Type of change

## How has this been tested?

Manually in the Sandbox

Given the defaults, nothing should change for any other ToggleButtonGroup, but good to check.

## Checklist

### DIFF
--- a/src/app/components/cds/Sandbox.js
+++ b/src/app/components/cds/Sandbox.js
@@ -197,6 +197,7 @@ const SandboxComponent = ({ admin }) => {
   const [toggleButtonGroupLabel, setToggleButtonGroupLabel] = React.useState(Boolean(true));
   const [toggleButtonGroupHelp, setToggleButtonGroupHelp] = React.useState(Boolean(true));
   const [toggleButtonGroupDisabled, setToggleButtonGroupDisabled] = React.useState(Boolean(false));
+  const [toggleButtonGroupisVertical, setToggleButtonGroupisVertical] = React.useState(Boolean(false));
 
   const [timeLabel, setTimeLabel] = React.useState(Boolean(false));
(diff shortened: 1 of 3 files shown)


brianfleming (changes requested):
nice, looking good @sarahkeh 
The only change I would request for this PR would be to add in the borders between the buttons when it goes to vertical. I think you just need to switch from left and right borders to top and bottom probably

amoedoamorim (approved):
Code LGTM, but please address Brian's requested changes

sarahkeh:
Added a border between buttons, lmk if it looks good @brianfleming

brianfleming (changes requested):
Thanks for that border change @sarahkeh looks good

Just noticed one other minor thing, if you toggle the disabled flag and the theme you will see that the "standard" theme disabled active button color is different than the new settings theme disabled color, we should just align that new theme just in case.

### this: correct disabled active bg color
Screenshot 2024-12-17 at 1 18 36 PM


### vs: still using the purple
Screenshot 2024-12-17 at 1 18 41 PM

Also, just a workflow thing, but hit the re-request review so I get pinged when you want me to look again

brianfleming (approved):
nice @sarahkeh thanks for all the refinements!

PR #2234, opened 2024-12-17
CV2-5748 enable filter for list, general cleanup of superAdminMask prop

sarahkeh:
## Description

Added check for localstorage contentMask value to AspectRatio.  This covers all content images not handled by the ItemThumbnail.  Removed superAdminMask prop being passed down the Media component tree.  Changed the default value for IgnoreGeneralContentMask to false (keeping in place, though, just in case there is a situation in which it could be useful) and removed value passed from the Item List.  Also updated the admin filter to first look for a localstorage value, then fall back to on if no value is found, as the admin mask currently works (will change to off in the next ticket).  This lets the user's choice persist throughout the app.

References: CV2-5748

## Type of change

## How has this been tested?

To turn the filter on for the list, currently you have to go to an item, ensure the filter is on, then go back to the list.  Similarly, if you turn the filter off, then navigate to another item or the list the filter should stay off.  

- Etc.

## Checklist

### DIFF
--- a/src/app/components/cds/media-cards/ItemThumbnail.js
+++ b/src/app/components/cds/media-cards/ItemThumbnail.js
@@ -132,7 +132,7 @@ function requiredBoolIfTypeNotEmpty(props, propName, componentName) {
 }
 
 ItemThumbnail.defaultProps = {
-  ignoreGeneralContentMask: true,
+  ignoreGeneralContentMask: false,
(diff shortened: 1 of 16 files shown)


amoedoamorim (approved):
Looks great! I'm glad we're getting rid of this much of prop drilling! :tada:

PR #2254, opened 2025-01-17
Hard-refresh when switching workspaces.

caiosba:
## Description

While debugging an error related to creating tipline resources, I realized it occurred only when navigating to the tipline settings page directly from the Workspaces list (i.e., switching from one workspace to another). The resource isn't created because the workspace ID being used is still the previous one. As a result, the session isn't fully cleared or switched to the new workspace. The simplest fix is to force a hard refresh to ensure a completely new session under the new workspace.

Fixes: [CV2-6003](https://meedan.atlassian.net/browse/CV2-6003).

## Type of change

## How has this been tested?

Manually. Just make sure you can still switch between workspaces.

## Checklist

[CV2-6003]: https://meedan.atlassian.net/browse/CV2-6003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/user/UserWorkspaces/PaginatedUserWorkspaces.js
+++ b/src/app/components/user/UserWorkspaces/PaginatedUserWorkspaces.js
@@ -65,6 +65,12 @@ const UserWorkspacesComponent = ({
   };
 
   const onSuccess = (team) => {
+    /* While debugging an error related to creating tipline resources, I realized it occurred only
+       when navigating to the tipline settings page directly from the Workspaces list (i.e., switching
+       from one workspace to another). The resource isn't created because the workspace ID being used
+       is still the previous one. As a result, the session isn't fully cleared or switched to the new
+       workspace. The simplest fix is to force a hard refresh to ensure a completely new session under
+       the new workspace. */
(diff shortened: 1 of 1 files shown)


brianfleming (approved):
womm

src/app/components/user/UserWorkspaces/PaginatedUserWorkspaces.js
```diff
@@ -66,8 +65,7 @@ const UserWorkspacesComponent = ({
   };
 
   const onSuccess = (team) => {
-    const path = `/${team.slug}/all-items`;
-    browserHistory.push(path);
+    window.location.assign(`/${team.slug}/all-items`);
```
amoedoamorim, line 68: Suggestion: Let's leave a comment mentioning why this hard reset is important, otherwise it might get inadvertently replaced with a `browserHistory.push` :sweat_smile:

caiosba: Great idea, @amoedoamorim! Done!

PR #2285, opened 2025-02-07
Fix TypeError: Cannot destructure property 'name' of '{}' as it is null.

danielevalverde:
## Description

Avoid trying to destructure the prop of null

References: 5510

## How to test?
Running unit tests and manually 

## Checklist

### DIFF
--- a/src/app/components/media/MediaOrigin.js
+++ b/src/app/components/media/MediaOrigin.js
@@ -121,7 +121,7 @@ const MediaOrigin = ({
     media_cluster_origin_user: user = {},
   } = {},
 }) => {
-  const { icon, message, tooltipMessage } = getIconAndMessage(origin, user?.name);
+  const { icon, message, tooltipMessage } = getIconAndMessage(origin, user.name);
 
   return (
     <Tooltip arrow title={tooltipMessage}>
--- a/src/app/components/media/MediaOrigin.js
+++ b/src/app/components/media/MediaOrigin.js
@@ -118,10 +118,10 @@ const getIconAndMessage = (origin, user) => {
 const MediaOrigin = ({
   projectMedia: {
     media_cluster_origin: origin,
-    media_cluster_origin_user: { name: user } = {},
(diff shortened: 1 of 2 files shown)


src/app/components/media/MediaOrigin.js
```diff
@@ -118,10 +118,10 @@ const getIconAndMessage = (origin, user) => {
 const MediaOrigin = ({
   projectMedia: {
     media_cluster_origin: origin,
-    media_cluster_origin_user: { name: user } = {},
-  },
+    media_cluster_origin_user: user = {},
+  } = {},
```
amoedoamorim, line 122: [comment] (non-blocker) I'm not a fan of this nested destructuring, but it works. Just a personal preference, I guess... :shrug:

src/app/components/media/MediaOriginBanner.js
```diff
@@ -109,11 +109,11 @@ const MediaOriginBanner = ({
   projectMedia: {
     media_cluster_origin: origin,
     media_cluster_origin_timestamp: originTimestamp,
-    media_cluster_origin_user: { name: user } = {},
+    media_cluster_origin_user: user = {},
     media_cluster_relationship: mediaClusterRelationship,
   },
 }) => {
-  const { icon, message } = getIconAndMessage(origin, mediaClusterRelationship, user, originTimestamp);
+  const { icon, message } = getIconAndMessage(origin, mediaClusterRelationship, user?.name, originTimestamp);
```
amoedoamorim, line 116: [comment] (non-blocker) If user is being assigned an empty object by default then the safe navigation operator shouldn't be necessary, right?

amoedoamorim (approved):

src/app/components/media/MediaOriginBanner.js
danielevalverde: removed the unnecessary safe operator, @amoedoamorim :saluting_face:

PR #2287, opened 2025-02-08
CV2-5958 Add custom slider component

sarahkeh:
## Description

Added a slider component that uses our custom tooltip.  Adds a marked prop which lets us move out the slider specific styling from the BotPreview css.  Added to Sandbox.

References: CV2-5958

## How to test?

Manually, no functional change

## Checklist

### DIFF
--- a/src/app/components/bot/BotPreview.module.css
+++ b/src/app/components/bot/BotPreview.module.css
@@ -187,48 +187,6 @@
 
       .settings-slider-wrapper {
         margin: 8px 24px;
-
-        :global(.MuiSlider-rail) {
-          border-radius: var(--border-radius-medium);
-          color: var(--color-gray-96);
-          height: 10px;
-        }
-
-        :global(.MuiSlider-track) {
-          display: none;
-        }
-
-        :global(.MuiSlider-mark) {
-          color: var(--color-gray-88);
-          height: 18px;
-          margin-top: -4px;
-          width: 3px;
-
-          &:hover {
-            border: 12px solid var(--color-blue-32);
-            border-radius: var(--border-radius-max);
-            margin-left: -10px;
-            margin-top: -7px;
-            opacity: .2;
-          }
-        }
-
-        :global(.MuiSlider-thumb) {
-          height: 22px;
-          margin-left: -10px;
-          margin-top: -7px;
-          width: 22px;
-        }
-
-        :global(.MuiSlider-markLabel) {
-          color: var(--color-gray-15);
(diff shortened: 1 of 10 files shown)


caiosba (changes requested):
Code looks good! I also tested it locally and it works well for:

- Bot preview
- Sandbox

My only requested change here is just to add unit tests for the `<SliderComponent />`, @sarahkeh... it's pretty self-contained, so it should be straightforward.

I also have a couple of questions, @sarahkeh: 

1. Did you name it `SliderComponent` instead of `Slider` because of name clash with MaterialUI's one? If so, what do you think about naming it `Slider` and importing Materia UI's one with something like `import { Slider as MuiSlider } from '@material-ui/core/Slider'`?
2. Did you check if using this new slider component is simple for these other cases? If not, please file a separate refactoring ticket for it:
  2.1. `src/app/components/cds/media-cards/MediaVolume.js`
  2.2. `src/app/components/cds/inputs/SliderComponent.js`
  2.3. `src/app/components/team/Similarity/ThresholdControl.js`

sarahkeh:
@caiosba I'll add a some tests

1. Yeah, the naming overlap is why I went with SliderComponent.  I also looked to the SwitchComponent, another wrapper we use for a MaterialUI component.  But I like your idea, it's a better match with the rest of the CDS components. 
2. I thought about that, wasn't sure if that was within the scope of this ticket.  If the update is as simple as changing the import, I'll add those, too, otherwise I'll make a separate ticket.

caiosba:
> @caiosba I'll add a some tests
> 
> 1. Yeah, the naming overlap is why I went with SliderComponent.  I also looked to the SwitchComponent, another wrapper we use for a MaterialUI component.  But I like your idea, it's a better match with the rest of the CDS components.
> 2. I thought about that, wasn't sure if that was within the scope of this ticket.  If the update is as simple as changing the import, I'll add those, too, otherwise I'll make a separate ticket.

Awesome, thanks Sarah!

sarahkeh:
@caiosba 

Updated the MediaVolume slider and the Similarity Threshold Controls with just an import change.  To make sure the slider looked the same, I added an extra check to the (now renamed) Slider component to not return the Tooltip when displayValueComponent is off.  Added two tests to verify the slider loads correctly with and without marks.

sarahkeh:
@caiosba Oh, good catch! Just replaced that one, too.

caiosba (approved):
Glad this last change was straightforward :)

GTG! :partying_face:

PR #2288, opened 2025-02-08
CV2-5781 remove save from Shared Feed

sarahkeh:
## Description

Shared feeds that displayed Media Clusters had an unusable Save button.  There was already a disableSave value being passed, the logic was just backwards, where if disableSave was true, the button would show.

References: CV2-5781

## How to test?

Set up a Shared Feed for Media Clusters and Requests, verify there's no Save button besides the filters

## Checklist

### DIFF
--- a/src/app/components/feed/FeedFilters.js
+++ b/src/app/components/feed/FeedFilters.js
@@ -1,4 +1,3 @@
-/* eslint-disable react/sort-prop-types */
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


src/app/components/feed/FeedFilters.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable react/sort-prop-types */
```
caiosba, line 1: Thanks for this!

caiosba (approved):
Thanks for the clean-up as well!

PR #2301, opened 2025-02-13
CV2 5687 removed unmatched media list

sarahkeh:
## Description

Removed unmatched media list page, sidebar link, related route and page props. 

References: CV2-5687

## How to test?

Manually, /unmatched-media no longer directs to a list and the list no longer appears in the sidebar. 

## Checklist


caiosba (approved):
Looks good! I also tested it locally and confirmed that there are no other instances of `unmatched-*` to be removed. All the repeated changes to the `page` prop here across different components reminds me of a refactoring I asked Alex to work on soon, which is to define some `PropTypes` that we can reuse across different components, like `articleType` and `coreLists`. I noted this to talk to Alex again, let's see if we can do it in March.

PR #2302, opened 2025-02-13
CV2-6161 link team to bot preview history

sarahkeh:
## Description

To keep chat history tied to specific workspaces, I added the team slug as a prefix to the botPreviewMessageHistory key.  Now multiple chat histories can be saved based on what workspace the user is in.  

References: CV2-6161

## How to test?

Since this changes the storage key for the chat history, it will be like starting a new chat.  Send a message or two in the chat, check localstorage to confirm there's a <team-slug>BotPreviewMessageHistory key, then switch to a different workspace and do the same thing.  Verify the two chats are different.  

## Checklist


caiosba (approved):
Tested locally and works well! It's fine to "reset" current histories since this feature is behind a feature flag.

PR #2311, opened 2025-02-21
Alert component adjustments

brianfleming:
## Description

While working on designs for CV2-5171 I needed some different variants of the `<Alert...` component. I decided to make these changes in the code at the same time so they would already be available. I also took the opportunity to do a bit of refactoring for the future:
- Added a `Placement` prop type of `oneOf` instead of individual props
- Added a `Border` prop in order to be able to toggle off the border individually. Also removed a classname override for media origin banner to take advantage of this prop.
- Updated the html so wherever there was a border style, the appropriate prop was added since `Border` defaults to `false`
- Updated the sandbox to include the new `Border` prop
- Sorted prop types when file touched

References: CV2-5171

## How to test?

There should be no visible changes, but this can be confirmed via the Sandbox where I added toggle for the new border prop type

## Checklist

### DIFF
--- a/src/app/components/BrowserSupport.js
+++ b/src/app/components/BrowserSupport.js
@@ -30,8 +30,8 @@ class BrowserSupport extends Component {
     if (BrowserSupport.shouldShowMessage()) {
       return (
         <Alert
-          banner
           icon
+          placement="banner"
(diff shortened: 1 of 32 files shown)


brianfleming:
Ran smoke tests on this branch, but I think I am running in to flaky tests instead of actual failures?

PR #2314, opened 2025-02-22
Cv2 5821 filter platforms on dashboard

sarahkeh:
## Description

By connecting to new statistics_platforms value, the tipline dashboard now only shows the platforms that have data associated with them.  If no platforms have data, a default message is shown.  

References: CV2-5821

## How to test?

Tipline dashboard -> Platforms: All dropdown should no longer show unused options.  If there is no data on any platform, a disabled option saying "No available platforms" should display 
## Checklist


src/app/components/dashboard/TiplineDashboard.js
```diff
@@ -113,6 +113,7 @@ TiplineDashboard.propTypes = {
     get_language: PropTypes.string.isRequired,
     get_languages: PropTypes.string.isRequired,
     slug: PropTypes.string.isRequired,
+    smooch_bot: PropTypes.object.isRequired,
```
amoedoamorim, line 116: [question] why is this line being added?

amoedoamorim: From your comment in the ticket, I'm inferring this was introduced but was no longer necessary. If so, please remove this line.

Also, can you please change the `data_report` propType to `array` instead of `object`? Not related to this ticket but I noticed a propType validation error while testing and it would be great to fix it in this PR! Thanks!

src/app/components/dashboard/PlatformSelect.js
```diff
@@ -21,15 +22,31 @@ const PlatformSelect = ({
...
-      Object.entries(humanTiplineNames).map(([k, v]) => (
-        <option key={k} value={String(k).toLowerCase()}>{v}</option>
-      ))
+    { platforms.length ?
+      Object.entries(humanTiplineNames).map(([k, v]) => {
+        if (platforms.indexOf(String(k).toLowerCase()) > -1) {
+          return <option key={k} value={String(k).toLowerCase()}>{v}</option>;
+        }
+        return null;
+      })
```
amoedoamorim, line 31: Maybe shuffling the logic a little would make this more readable? Perhaps something like `platforms.map(p => <option key={p} value={p}>{humanTiplineNames[p]}</option>)` would work?

amoedoamorim (approved):
Looks good. I'm leaving a suggestion for improvement on readability.

src/app/components/dashboard/PlatformSelect.js
sarahkeh: Good call, just pushed a little cleanup to only use humanTIplineNames for the label.

PR #2321, opened 2025-02-28
Fixing how URL parameter is set when an article is clicked from the "All Articles" list.

caiosba:
## Description

Make sure that in the URL we have either `factCheckId` or `explainerId` in the URL, never both.

Reference: CV2-5007.

## How to test?

Manually: From the "All Articles" list, click on different articles, and make sure that the right article is loaded in the slideout and that the URL contains either an `explainerId` or `factCheckId`, never both.

## Checklist


danielevalverde (approved):
:LGTM:

PR #2332, opened 2025-03-18
Refactoring/CV2-5203: Return all items that match a search

danielevalverde:
## Description

Previously, when a fact-check was linked to another item, it was excluded from search results. We modified the query parameter to return all media types to ensure that items appear in search results along with their associated fact-checks. For the fact-check tab, it was only returning items of blank type:
[AutoCompleteMediaItem.js#L105](https://github.com/meedan/check-web/blob/develop/src/app/components/media/AutoCompleteMediaItem.js#L105).

Additionally, added a tooltip to inform users that the fact-check is not applicable and disabled the item to prevent selection.

Reference: CV2-5203

## How to test?

- [ ] Go to an item page and click on "Merge media cluster".
- [ ] Navigate to the "Add to imported Fact Check" tab.
- [ ] Search for a fact-check that is already in use:
- [ ] Verify that it is not clickable.
- [ ] Hover over it and check if a tooltip appears, explaining that the fact-check is already in use.
- [ ] Search for a fact-check that is not in use:
- [ ] Verify that it is selectable.
- [ ] Select it and confirm that it can be added to the item successfully.

## Checklist

### DIFF
--- a/localization/react-intl/src/app/components/media/AutoCompleteMediaItem.json
+++ b/localization/react-intl/src/app/components/media/AutoCompleteMediaItem.json
@@ -43,5 +43,10 @@
     "id": "autoCompleteMediaItem.cantSelectPublished",
     "description": "Tooltip error message about when media can be imported",
     "defaultMessage": "Media cannot be imported from items that have their report published"
+  },
+  {
+    "id": "autoCompleteMediaItem.factcheckInUseTooltip",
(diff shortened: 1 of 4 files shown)


caiosba (approved):
Great work, Dani! Reflects what I expected from our "pairing". Thanks for the clean-up (`typesToShow` prop, etc.) and for the test plan in the PR.

PR #2346, opened 2025-04-14
Webhook Management UI

amoedoamorim:
## Description

Creates the new Webhook management UI

References: CV2-4716

## How to test?

- As a super admin, go to workspace settings > Integrations > External Webhooks and try, creating, updating and deleting a webhook

## Checklist


amoedoamorim:
Asked for Manu's review just for running locally and trying the UI

vasconsaurus (approved):
Tested locally, and it looks good 🎉 

My only note is: when I deleted a webhook, I had to refresh the page, and open the webhook tab to confirm the webhook was deleted.

src/app/components/team/WebhookEntry.js
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+      </div>
+    </div>
+  );
+};
+
+WebhookEntry.propTypes = {
+  webhook: PropTypes.shape({
+    name: PropTypes.string,
+    events: PropTypes.arrayOf(PropTypes.shape({
```
caiosba, line 48: I think we can replace `PropTypes.shape` by `PropTypes.exact` here, no?

src/app/components/team/Webhooks.js
```diff
@@ -0,0 +1,141 @@
...
+        </div>
+      </Collapse>
+    </div>
+  );
+};
+
+Webhooks.propTypes = {
+  team: PropTypes.shape({
+    webhooks: PropTypes.shape({
+      edges: PropTypes.arrayOf(PropTypes.object.isRequired).isRequired,
```
caiosba, line 124: Actually here could be something like:

```javascript
edges: PropTypes.arrayOf(
  PropTypes.exact({
    node: PropTypes.shape({
     dbid: PropTypes.number.isRequired,
    }).isRequired,
  }).isRequired,
).isRequired,
```

Right?

caiosba (changes requested):
Very good, thanks @amoedoamorim - I left a few comments, mostly related to some `propTypes`.

The issue that @vasconsaurus mentioned about deleting doesn't occur in production mode, so it's likely related to that [Relay Compat issue that happens only in development mode](https://github.com/meedan/check-web?tab=readme-ov-file#notes-and-tips). So, no worries there.

However, I noticed a problem while testing locally: I was able to successfully create a webhook with only a name (even though those other fields are supposed to be mandatory). Since it doesn't have any events or a request URL, it's not listed afterward. I think we should add client-side validation for this.

Last thing: Please add unit tests for the new components, even if they just make sure the components are mounted correctly.

src/app/components/team/WebhookEdit.js
```diff
@@ -0,0 +1,394 @@
...
+  const setFlashMessage = React.useContext(FlashMessageSetterContext);
+
+  const resetForm = () => {
+    setName('');
+    setHeaders('');
+    setUrl('');
+    setEventType('update_annotation_verification_status');
+    setNameError(false);
+    setUrlError(false);
+    setHeadersError(false);
```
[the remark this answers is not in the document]
amoedoamorim: Hmm, I guess it could, I'm just not sure it should. Is there any benefit other than less individual states?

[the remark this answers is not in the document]
caiosba: Yes, the benefit is more atomic updates, instead of one update for each `set*`. But not big deal for light UIs like this.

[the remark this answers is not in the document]
caiosba: It depends... since our fields are logically related (e.g., part of a form), using a single useState with an object can improve readability, maintainability, and reset logic. But keeping separate states allows more precise control and avoids unnecessary re-renders.

src/app/components/team/Webhooks.js
amoedoamorim: `PropTypes.exact` doesn't seem to work quite well with graphql fragment props, as they might contain additional metadata keys which break the exact match:

image

I'm keeping this as is, as I don't think we need to be too specific here. `PropTypes.object` should be enough when the actual shape is defined in the data fragment.

caiosba: Good point on the meta-fields! Agreed.

caiosba (approved):
Thanks Alelas for going through the comments from the previous review. Everything works well locally too!

PR #2348, opened 2025-04-17
Fix/CV2-6262: Do not display Messenger template messages in chat history

danielevalverde:
## Description

- [X]  Filter Messenger template messages (tag: ACCOUNT_UPDATE and message_type: MESSAGE_TAG) sent after 24h from being displayed in the chat, in order to avoid duplicates and unsupported messages
- [X] Add unit test in Messenger.test.js to reproduce and cover the issue
- [X]  Add unit test in ChatFeed.test.js to ensure the solution works as expected

References: CV2-6262

## How to test?

- Run the unit tests
- Manually check the chat history in the UI to ensure that unsupported or duplicate Messenger template messages are no longer shown after the 24h window

## Checklist

### DIFF
--- a/src/app/components/cds/chat/ChatFeed.js
+++ b/src/app/components/cds/chat/ChatFeed.js
@@ -41,6 +41,11 @@ const ChatFeed = ({
   const parseItem = (item) => {
     let output = '';
 
+    // Skip Messenger template messages sent outside the 24h window to avoid duplicates and unsupported messages in the chat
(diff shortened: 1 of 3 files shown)


caiosba (approved):
Very good, thanks Dani!

caiosba:
I forgot to mention that I tested it locally - I encountered some cases of those "unsupported message" related to reports, and they are gone after this fix, while the valid messages are still being displayed.

PR #2364, opened 2025-05-22
5217, 6373 – Shared Feed: refactoring, copy, and design updates

vasconsaurus:
## Description
This works updates the creation of a shared feed, so we are able to select and save a media list and/or an article list. 

This only updates the creation of the feed, the display of both lists in the Feed will be a separate ticket. This should not break the current Feed view.

### 1. On design changes
We are following the new design in [jira](https://www.figma.com/file/N1W1p7anE8xxekD7EyepVE?type=design&node-id=9367%3A4072). So that article and media `FeedDataPointsSection` both have their own `FeedContent` component, where only article lists or media lists will be displayed, we had to:
- update how we structure the components,
- update some components to accept a `listType` prop,
- update copy in the components

**Structure: Before**
We had the two FeedDataPointsSection, inside FeedDataPoints, which was inside SaveFeed. While FeedContent was inside SaveFeed.  

Screenshot 2025-06-10 at 14 11 16

**Structure: After**
I moved FeedContent into FeedDataPointsSection, which was moved directly into SaveFeed. I had to add a new FeedDataPointsSectionHeader with the title and alert.

Screenshot 2025-06-10 at 14 11 05

### 2. Article saved search
Currently, we only save one reference to a `media_saved_search` in shared feeds. We are updating the backend, concurrently to this work (https://github.com/meedan/check-api/pull/2310), to be able to save one reference to an `article_saved_search` as well.

We updated so we are able to:
- select two lists if needed (one for media and one for articles)
- query, create, update, delete both lists

### 3. Videos

**3.1 Creating a new feed**
https://github.com/user-attachments/assets/82b89e91-ad20-4bef-ab6f-faeff624f1bc

**3.1 Accepting an invitation to a new feed**
https://github.com/user-attachments/assets/f0baef18-6989-4d72-a22d-faf962a4003c

**3.3 Displaying and updating an existing feed**
https://github.com/user-attachments/assets/35a2a137-efa5-4692-92d3-9b838bcceb5a

References: CV2-5217, CV2-6373

## How to test?
Manually, try to create a new shared feed, click the toggles and try to select either of the lists, and create a new feed. 

## Checklist

### DIFF
--- a/localization/react-intl/src/app/components/feed/FeedDataPointsSection.json
+++ b/localization/react-intl/src/app/components/feed/FeedDataPointsSection.json
@@ -10,11 +10,6 @@
     "defaultMessage": "Media Clusters and Requests"
   },
   {
-    "id": "feedDataPoints.title",
-    "description": "Title of section where the data points being shared in a feed can be selected.",
-    "defaultMessage": "What workspace data will be shared in this feed?"
-  },
-  {
     "id": "feedDataPoints.ArticleContentTitle",
     "description": "Title of the content section on save feed page where a user can select that fact-checks will be shared in this feed.",
     "defaultMessage": "<strong>Article</strong> data to be shared:"


package.json
danielevalverde, line 1: Manu, Could you please revert this change and remove the additional --fix flag?  It's generally better to avoid using --fix in the linter script to ensure that fixes are applied intentionally and reviewed before committing.

danielevalverde (commented):
Thank you for sharing the draft! The code structure looks good, I just left a comment on the PR. As most of the implementation code was already reviewed and discussed during the pairing sessions.

But I noticed an issue: when selecting a list for articles or media in the dropdown, the other dropdown is being reset and does not persist the previously selected value. I believe this is happening because both dropdowns are using the same state (selectedListId).

To fix this, I suggest separating the states for each dropdown instead of using a single selectedListId. This way, each dropdown can manage its selection independently without interfering with the other.

package.json
vasconsaurus: My intention was to only keep it temporarily, but I could have forgotten.

vasconsaurus:
> Thank you for sharing the draft, @vasconsaurus ! The code structure looks good, I just left a comment on the PR. As most of the implementation code was already reviewed and discussed during the pairing sessions.
> 
> But I noticed an issue: when selecting a list for articles or media in the dropdown, the other dropdown is being reset and does not persist the previously selected value. I believe this is happening because both dropdowns are using the same state (selectedListId).
> 
> To fix this, I suggest separating the states for each dropdown instead of using a single selectedListId. This way, each dropdown can manage its selection independently without interfering with the other.

@danielevalverde a question: right now we can only save one saved search in the backend side. We can either pick an article saved search or a media clusters saved search. My next backend ticket will be updating that, so we can store the two saved searches. 

So I'm wondering if this isn't the current behavior we want? What do you think?

danielevalverde:
> > Thank you for sharing the draft, @vasconsaurus ! The code structure looks good, I just left a comment on the PR. As most of the implementation code was already reviewed and discussed during the pairing sessions.
> > But I noticed an issue: when selecting a list for articles or media in the dropdown, the other dropdown is being reset and does not persist the previously selected value. I believe this is happening because both dropdowns are using the same state (selectedListId).
> > To fix this, I suggest separating the states for each dropdown instead of using a single selectedListId. This way, each dropdown can manage its selection independently without interfering with the other.
> 
> @danielevalverde a question: right now we can only save one saved search in the backend side. We can either pick an article saved search or a media clusters saved search. My next backend ticket will be updating that, so we can store the two saved searches.
> 
> So I'm wondering if this isn't the current behavior we want? What do you think?

I believe this ticket might need to be paused until the backend ticket is completed, as it will allow us to store the two saved searches and avoid this behavior. However, I think it would be best to first confirm the expected behavior with 
 @brianfleming, @vasconsaurus, before moving forward.

src/app/components/feed/FeedContent.js
```diff
@@ -68,12 +55,12 @@ const FeedContent = ({
...
 FeedContent.defaultProps = {
   listId: null,
-  dataPoints: [],
+  listType: null,
 };
 
 FeedContent.propTypes = {
   listId: PropTypes.number,
-  dataPoints: PropTypes.arrayOf(PropTypes.number.isRequired),
+  listType: PropTypes.string,
```
amoedoamorim, line 63: What are the possible values for `listType`?
Perhaps something like `listType: PropTypes.oneOf(['media', 'article'])` would be a better propType definition

amoedoamorim: Suggestion: renaming this component `FeedContent` => `FeedSelectList` to better reflect what it does

src/app/components/feed/FeedDataPointsSection.js
```diff
@@ -34,26 +51,186 @@ const FeedDataPointsSection = ({
...
+  listId: null,
+  readOnly: false,
   onToggle: () => {},
 };
 
 FeedDataPointsSection.propTypes = {
-  content: PropTypes.node.isRequired,
   enabled: PropTypes.bool,
+  listId: PropTypes.number,
+  listType: PropTypes.string.isRequired,
```
amoedoamorim, line 230: What are the possible values for listType?  
Perhaps something like `listType: PropTypes.oneOf(['media', 'article'])` would be a better propType definition

amoedoamorim (approved):
:lgtm:

Just added a suggestion to improve `listType` propTypes

danielevalverde (approved):
:LGTM:

PR #2377, opened 2025-06-24
6419 – Do not show NavigateAwayDialog if saving

vasconsaurus:
## Description

When saving as an invited team, the following notice would pop-up:
'Do you want to leave without saving? You have unsaved changes to your shared feed. …'

This happened because even though we called handleSaveFeedTeam successfully,
and the feed was saved, the `isEditing` state was not set to false.
Which caused NavigateAwayDialog to pop up.

I think as part of the save operation, we should also set `isEditing` to false,
if someone is saving, they are no longer editing.

References: CV2-6419

## How to test?

- As user 1, create a shared feed and invite user 2
- As user 2:
    - accept the invitation,
    - add a saved search,
    - click save
- The user should be redirected to the Feed page

## Checklist

### DIFF
--- a/src/app/components/feed/SaveFeed.js
+++ b/src/app/components/feed/SaveFeed.js
@@ -372,6 +372,7 @@ const SaveFeed = (props) => {
 
   const handleSaveFeedTeam = () => {
     setSaving(true);
+    setIsEditing(false);
     const input = {
       id: feedTeam.id,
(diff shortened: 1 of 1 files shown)


vasconsaurus:
@danielevalverde, if this makes sense, do you think I should add a test?

caiosba:
@danielevalverde , can you please take this review

PR #2401, opened 2025-08-04
Support legacy deployments via sysops

sonoransun:
## Description

This change adds support for legacy deployments via sysops using GitHub Actions workflow.

References: CV2-6471

## Checklist


sonoransun:
@danielevalverde this is ready for testing on the sysops side, but this needs to be merged first. All builds should continue as normal, but I do expect the last step of the deploy into sysops to maybe fail. TBD

danielevalverde (approved):

PR #2412, opened 2025-08-20
CV2-4862: Prevent crash on Home page when user.user_teams is undefined or invalid JSON

danielevalverde:
## Description

Add a safe JSON parse for user.user_teams in HomeComponent to avoid runtime errors when the value is undefined or not valid JSON.

References: CV2-4862

## Checklist

### DIFF
--- a/src/app/components/Home.js
+++ b/src/app/components/Home.js
@@ -19,7 +19,7 @@ import UserTos from './UserTos';
 import CheckContext from '../CheckContext';
 import { withClientSessionId } from '../ClientSessionId';
 import { stringHelper } from '../customHelpers';
-import { bemClass } from '../helpers';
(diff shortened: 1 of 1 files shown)


src/app/components/Home.js
```diff
@@ -265,7 +265,7 @@ class HomeComponent extends Component {
         if (user.is_admin) {
           return true;
         }
-        const teams = JSON.parse(user.user_teams);
+        const teams = typeof user?.user_teams === 'string' ? JSON.parse(user.user_teams) : {};
```
amoedoamorim, line 268: why not just use our `safelyParseJSON` helper?

danielevalverde: Sorry @amoedoamorim,  I forgot we already have this helper. I've just pushed the update :)

